PR #14, opened 2022-04-17
added huber loss and other schedulers

kashif:


lucidrains:
brilliant! 🙏❤️

kashif:
thank you!

PR #16, opened 2022-04-20
Fix denoising sampling

kashif:
The `q_posterior` function as implemented is not correct and is not needed for sampling. Recall in Ho et al. equation (7) is for `x_0` which is not known during sampling.

And in Alg. 2 line 4 we just need the slightly denoised sample.

Issue flagged to me by @NielsRogge so all credit to him if I am not mistaken... and if I am mistaken then it's my fault!


NielsRogge:
Disclaimer: I'm still figuring out the sampling code :D not sure what's correct. I'm just wondering to which equations in the DDPM paper the `predict_start_from_noise` and `q_posterior` methods correspond.

kashif:
closing the PR as https://github.com/hojonathanho/diffusion/issues/5 seems to answer it...

PR #17, opened 2022-04-20
added diffusion-gan thoughts

kashif:
https://github.com/NVlabs/denoising-diffusion-gan


lucidrains:
love it! i like the lightweight gan idea :) https://github.com/lucidrains/lightweight-gan

Issue #23, opened 2022-04-24
open replication of the prior


rom1504:
hey,

thanks @lucidrains for building this awesome replication of the model, as usual!
dalle2 paper https://arxiv.org/pdf/2204.06125.pdf

with a few people from laion we're working on replication of the prior at scale
we're gathering notes in https://docs.google.com/document/d/1BKIQPzZS7pVL2JgL74W0dUIlUcfld8jptA6nIne_cNo/edit
and here
big plan:
* [x] use https://github.com/rom1504/embedding-reader to build a data loader
  * [x] use laion2B-en 2B image/text embeddings https://mystic.the-eye.eu/public/AI/cah/laion5b/embeddings/laion2B-en/img_emb/
* [x] plug that into the diffusion prior here: need to support precomputed embeddings
* [x] train at small scale
* [x] implement metrics (eg retrieval, classification, generation) on the output of the prior
* [x] train at large scale

Anybody interested in that project, feel free to discuss here or in #dalle2-prior at laion server (https://discord.gg/xBPBXfcFHd)

hp param from dalle2 paper
![image](https://user-images.githubusercontent.com/2346494/164995465-79f1e6e6-5aba-415f-918d-e349ac5119cd.png)

As a first step, we're trying to have an end to end version running at small scale. Once it works, we'll scale it up.

We intend to send PR in this repo for any improvement that seems worthwhile (eg support precomputed embeddings, distributed training, ...)

lucidrains:
@rom1504 ah looks like a great plan :) so i think the PCA portion was only for the autoregressive prior method, and not the diffusion prior (they needed to quantize so they can use the straightforward cross entropy loss). but correct me if i'm wrong i'm happy to build that into the framework

rom1504:
ohhh you're right, well that makes thing much easier
I'll remove that from the plan

rom1504:
I'll keep this issue updated as we make progress

lucidrains:
![6dp8bj](https://user-images.githubusercontent.com/108653/164996908-097ea317-4928-4bb8-a3a1-434a1eb46fb2.jpg)

nousr:
I was able to over-fit a small subset of LAION-2B locally (using the new CLIP-less DiffusionPrior class). I'll be working to make that data loader next so we can do the first "real" training run. 

_referencing_:
>  - [ ] use https://github.com/rom1504/embedding-reader to build a data loader
[...]

PR #24, opened 2022-04-26
Add Functionality For 'Embedding-Only' Training

nousr:
I thought I would just get the ball rolling on this front! Feel free to suggest changes or comment on errors in my code.

The main motivation for this pull request is outlined in #23.

After some discussion in the LAION discord, it was proposed to try a framework that makes the following major changes:
- `DiffusionPrior` can now be instantiated with `clip=None` to specify embedding-only training.
- `DiffusionPrior.forward()` now has two added parameters that default to None: `text_embed` and `image_embed`

I don't yet have access to the text embeddings from LAION2B, however, I tested on dummy data of the same shape and was able to run with no errors thrown--this example usage is included below.

### DIFF
--- a/dalle2_pytorch/dalle2_pytorch.py
+++ b/dalle2_pytorch/dalle2_pytorch.py
@@ -423,54 +423,77 @@ class DiffusionPriorNetwork(nn.Module):
         *,
         text_encodings,
         text_embed,
-        mask = None,
-        cond_drop_prob = 0.2
+        mask=None,
+        cond_drop_prob=0.1
     ):
-        batch, text_enc_len, device = image_embed.shape[0], text_encodings.shape[-2], image_embed.device
+        """
+        Complete one forward pass through the diffusion prior network.
 
-        # in section 2.2, last paragraph
-        # "... consisting of encoded text, CLIP text embedding, diffusion timestep embedding, noised CLIP image embedding, final embedding for prediction"
+        If the dictionary passed from the DiffusionPrior wrapper has the 
+        key:value pair text_encodings:None, then we will assume
+        we are doing "embedding only" training.
 
-        text_embed, image_embed = rearrange_many((text_embed, image_embed), 'b d -> b 1 d')
+        In the case of embedding only training we will drop the text
+        encoding from the tokens entirely and train on the remaining
+        tokens mentioned in section 2.2 of the DALLE2 paper.
+        """
 
-        # whether text embedding is used for conditioning depends on whether text encodings are available for attention (for classifier free guidance, even though it seems from the paper it was not used in the prior ddpm, as the objective is different)
-        # but let's just do it right
+        # grab some contextual information
+        batch, device = image_embed.shape[0], image_embed.device
 
-        if exists(mask):
-            not_all_masked_out = mask.any(dim = -1)
-            mask = torch.cat((mask, rearrange(not_all_masked_out, 'b -> b 1')), dim = 1)
-
-        if exists(mask):
-            mask = F.pad(mask, (0, 2), value = True) # extend mask for text embedding, noised image embedding, time step embedding, and learned query
-
+        # gather tokens for casual transformer
+        text_embed, image_embed = rearrange_many(
+            (text_embed, image_embed), 'b d -> b 1 d')
         time_embed = self.time_embeddings(diffusion_timesteps)
         time_embed = rearrange(time_embed, 'b d -> b 1 d')
+        learned_queries = repeat(self.learned_query, 'd -> b 1 d', b=batch)
 
-        learned_queries = repeat(self.learned_query, 'd -> b 1 d', b = batch)
+        # check for embedding only training
+        if text_encodings is not None:
+            text_enc_len = text_encodings.shape[-2]
 
-        tokens = torch.cat((
-            text_encodings,
-            text_embed,
-            time_embed,
-            learned_queries
-        ), dim = -2)
+            # text conditioning can be passed during training for CFG; as described in the paper
 
-        # mask if it doesn't exist
+            if exists(mask):
+                not_all_masked_out = mask.any(dim=-1)
+                mask = torch.cat(
+                    (mask, rearrange(not_all_masked_out, 'b -> b 1')), dim=1)
+                # extend mask for text embedding, noised image embedding, etc.
+                mask = F.pad(mask, (0, 2), value=True)
 
-        if not exists(mask):
-            mask = torch.ones((batch, text_enc_len), device = device, dtype = torch.bool)
+            if not exists(mask):
+                mask = torch.ones((batch, text_enc_len),
+                                  device=device, dtype=torch.bool)
 
-        # classifier free guidance
+            # classifier free guidance
+            cond_prob_mask = prob_mask_like(
+                (batch,), cond_drop_prob, device=device)
+            mask &= rearrange(cond_prob_mask, 'b -> b 1')
 
-        cond_prob_mask = prob_mask_like((batch,), cond_drop_prob, device = device)
-        mask &= rearrange(cond_prob_mask, 'b -> b 1')
+            # actually combine everything as described originally
+            tokens = torch.cat((
+                text_encodings,
+                text_embed,
+                time_embed,
+                learned_queries
+            ), dim=-2)
 
-        # attend
+        # TODO: someone should double check this block of code below
+        else:
+            # during embedding only, just drop text_encodings entirely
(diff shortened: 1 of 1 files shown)


dalle2_pytorch/dalle2_pytorch.py
```diff
@@ -423,54 +423,77 @@ def forward(
...
-        # but let's just do it right
-
-        if exists(mask):
-            not_all_masked_out = mask.any(dim = -1)
-            mask = torch.cat((mask, rearrange(not_all_masked_out, 'b -> b 1')), dim = 1)
-
-        if exists(mask):
-            mask = F.pad(mask, (0, 2), value = True) # extend mask for text embedding, noised image embedding, time step embedding, and learned query
+        # grab some contextual information
+        batch, device = image_embed.shape[0], image_embed.device
```
nousr, line 442: pull out batch and device in the case of there being no text_encoding

dalle2_pytorch/dalle2_pytorch.py
```diff
@@ -423,54 +423,77 @@ def forward(
...
 
+        # gather tokens for casual transformer
```
nousr, line 444: gathered all tokens to the same chunk of code; leaving the text_encoding which is now considered optional.

dalle2_pytorch/dalle2_pytorch.py
```diff
@@ -423,54 +423,77 @@ def forward(
...
+        text_embed, image_embed = rearrange_many(
+            (text_embed, image_embed), 'b d -> b 1 d')
         time_embed = self.time_embeddings(diffusion_timesteps)
         time_embed = rearrange(time_embed, 'b d -> b 1 d')
+        learned_queries = repeat(self.learned_query, 'd -> b 1 d', b=batch)
+
+        # check for embedding only training
```
nousr, line 451: I opted to use a big if-statement block here to separate out the two possible training methods. I'm definitely open to more concise suggestions!

dalle2_pytorch/dalle2_pytorch.py
```diff
@@ -679,15 +710,31 @@ def sample(self, text, num_samples_per_batch = 2):
...
-        check_shape(image, 'b c h w', h = img_size, w = img_size, c = self.channels)
+    def forward(self, text, image, text_embed=None, image_embed=None, *args, **kwargs):
+        # check if we are doing embedding-only training
+        if text is not None:
+            b, device, img_size, = image.shape[0], image.device, self.image_size
+            check_shape(image, 'b c h w', h=img_size,
+                        w=img_size, c=self.channels)
+            text_cond = self.get_text_cond(text)
+        else:
+            b, device = image_embed.shape[0], image_embed.device
```
nousr, line 721: The DiffusionPriorNetwork expects a dictionary. So I format the predefined text embedding in a dictionary with text_encodings:None to make it possible to detect the embedding-only training once inside the DiffusionPriorNetwork.

dalle2_pytorch/dalle2_pytorch.py
```diff
@@ -423,54 +423,77 @@ def forward(
...
+            ), dim=-2)
+
+        # TODO: someone should double check this block of code below
+        else:
+            # during embedding only, just drop text_encodings entirely
+            tokens = torch.cat((
+                text_embed,
+                time_embed,
+                learned_queries
+            ), dim=-2)
```
nousr, line 488: I'd like someone to check this part (above) and verify that we can simply use a mask of all ones in the shape of `(batch, tokens.shape[1])`.

My understanding is that the mask is only used for the classifier-free guidance training (text_encoding). But I could be wrong!

nousr (commented):
I went through and added some comments on the changes I made to hopefully give some context for the problem and thought process for the changes.

dalle2_pytorch/dalle2_pytorch.py
```diff
@@ -423,54 +423,77 @@ def forward(
...
 
+        # gather tokens for casual transformer
+        text_embed, image_embed = rearrange_many(
+            (text_embed, image_embed), 'b d -> b 1 d')
         time_embed = self.time_embeddings(diffusion_timesteps)
         time_embed = rearrange(time_embed, 'b d -> b 1 d')
+        learned_queries = repeat(self.learned_query, 'd -> b 1 d', b=batch)
+
+        # check for embedding only training
+        if text_encodings is not None:
```
nousr, line 452: I opted to use a big if-statement block here to separate out the two possible training methods. I'm definitely open to more concise suggestions!

dalle2_pytorch/dalle2_pytorch.py
```diff
@@ -423,54 +423,77 @@ def forward(
...
-            mask = F.pad(mask, (0, 2), value = True) # extend mask for text embedding, noised image embedding, time step embedding, and learned query
+        # grab some contextual information
+        batch, device = image_embed.shape[0], image_embed.device
 
+        # gather tokens for casual transformer
+        text_embed, image_embed = rearrange_many(
+            (text_embed, image_embed), 'b d -> b 1 d')
         time_embed = self.time_embeddings(diffusion_timesteps)
         time_embed = rearrange(time_embed, 'b d -> b 1 d')
+        learned_queries = repeat(self.learned_query, 'd -> b 1 d', b=batch)
```
nousr, line 449: gathered all tokens to the same chunk of code; leaving the text_encoding which is now considered optional.

nousr (commented):
updated some comments

dalle2_pytorch/dalle2_pytorch.py
lucidrains: yup this looks good! 💯

dalle2_pytorch/dalle2_pytorch.py
lucidrains: yup, that's fine, another way would be to set text encodings to an empty tensor with the dimension at which to be concatenated to be 0

nousr:
> I tested on dummy data of the same shape and was able to run with no errors thrown--this example usage is included below.

```python

"""
Embedding-Only Training: Proof of Concept
"""

import torch
from dalle2_pytorch import DiffusionPriorNetwork, DiffusionPrior

# grab our models
prior_network = DiffusionPriorNetwork(
    dim=768,
    depth=6,
    dim_head=64,
    heads=8
).cuda()

# specify clip=None to signal for "embedding-only" training
diffusion_prior = DiffusionPrior(
    net=prior_network,
    clip=None,
    timesteps=64,
    cond_drop_prob=1
).cuda()

# LAION embeddings are (b, 768)
txt_embed = torch.normal(0, 1, (10, 768)).cuda()
img_embed = torch.normal(0, 1, (10, 768)).cuda()

# quick boilerplate 'training' to sanity check
loss = 0.0
EPOCHS = 10
optimizer = torch.optim.Adam(diffusion_prior.parameters())

for e in range(EPOCHS):

    optimizer.zero_grad()

    loss = diffusion_prior(text=None, image=None,
                           text_embed=txt_embed, image_embed=img_embed)
    loss.backward()
    loss = loss.item()
    optimizer.step()

    print(f"Epoch {e} | Loss: {loss}")

print(f"Test Completed!")```

dalle2_pytorch/dalle2_pytorch.py
lucidrains: yea, so here i'd actually just do `def forward(self, text = None, image = None, text_embed = None, image_embed = None)`

and later on do asserts to make sure one or the other is passed in `assert exists(text) ^ exists(text_embed)` and `assert exists(image) ^ exists(image_embed)` and handle them both in separate conditionals, prepping the kwargs for the `p_losses` function below

lucidrains:
love it! let's get this merged tomorrow morning (i'm off to 💤 )

lucidrains:
@nousr hey, i'm really sorry for doing this, but i think i'm going to hijack this and just knock out the code in the next hour or two. because you plan on training the embeddings alone (and it seems like you are also omitting text encodings), i will probably have the researcher specify upfront whether text encodings will be used. this will also change the logic for how i do the classifier free guidance (which is off the mask of the text encodings). and the third complication would be to account for all this when sampling

valiant efforts otherwise! glad to have someone else looking at the code with me!

nousr:
@lucidrains No worries! Thanks for taking the time to give some feedback. I look forward to the updates 🚀

lucidrains:
@nousr ok, let me know if this works for you! https://github.com/lucidrains/DALLE2-pytorch/tree/0.0.48#training-on-preprocessed-clip-embeddings (open up a new github issue if it does not)

Issue #29, opened 2022-04-28
Build a fair evaluation of the prior


rom1504:
We're starting to have our first decent prior now. (PR of the training script coming soon)

Time to evaluate
Ideas:
* Mse on test set
* Zero shot eval on image net Class Text -> text emb -> image emb -> ranking https://github.com/LAION-AI/project-menu/issues/13
* Clip guided generation with the prior

If you have more ideas please share, I may be missing some obvious things.

We have more volunteers that want to help so I'll point some here :)

lucidrains:
Already?! 🙏💯🎉

lucidrains:
I was going to add training scripts for all the components tomorrow morning ... 😂

lucidrains:
Eager to figure out just how important this prior network really is 🤔

lucidrains:
I think for the clip guided generation, you will still need a decoder conditioned on the clip image embedding, tho you can probably get away with a small resolution net for starters, just to validate

xiankgx:
Thank you for your work! I've modified your code to work with the original CLIP and got the training scripts. What is a good small dataset to test this on?

lucidrains:
@xiankgx 👋 are you working with Laion? you should, because they have a humongous (and smaller test) datasets

lucidrains:
> Thank you for your work! I've modified your code to work with the original CLIP and got the training scripts. What is a good small dataset to test this on?

Do you have a link to the CLIP that you used? I can try to incorporate it tomorrow using https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/dalle2_pytorch.py#L95

xiankgx:
> > Thank you for your work! I've modified your code to work with the original CLIP and got the training scripts. What is a good small dataset to test this on?
> 
> Do you have a link to the CLIP that you used? I can try to incorporate it tomorrow using https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/dalle2_pytorch.py#L95

I am using the CLIP from this link: https://github.com/openai/CLIP

lucidrains:
ohh got it, looks like they finally got it to be pip installable. I'll take a look tomorrow at an adapter!

lucidrains:
ok, the plan will be to automatically use the openai clip by setting a `use_openai_clip` on both the `Decoder` and `DiffusionPrior`, which will allow researchers to skip the first step in the whole process

xiankgx:
[...]

PR #34, opened 2022-04-29
Train DiffusionPrior with pre-computed embeddings

krish240574:
This is in response to https://github.com/lucidrains/DALLE2-pytorch/issues/29 - more metrics will get added.


dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+    for e in range(epochs):
+        train_loss = 0.0 
+        print("Training loop - epoch number ",e)
+        train_set_size = int(train_percent*image_reader.count)
+        for emb_images,emb_text in zip(image_reader(batch_size=batch_size, start=0, end=train_set_size),
+                text_reader(batch_size=batch_size, start=0, end=train_set_size)):
+            emb_images = list(emb_images)
+            emb_text = list(emb_text)
+            print(emb_images[0].shape,emb_text[0].shape)
+            if torch.cuda.is_available():
```
lucidrains, line 50: since you already set the device based on this logic above, do you need this again?

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+    parser.add_argument("--test-percent", type=float, default=0.1)
+    # LAION-style training(pre-computed embeddings)
+    parser.add_argument("--condition-on-text-encodings", type=bool, default=False)
+
+    args = parser.parse_args()
+    print("Setting up wandb logging... Please wait...")
+    wandb.init(
+      entity=args.wandb_entity,
+      project=args.wandb_project,
+      name=f"laion-dprior",
```
lucidrains, line 130: anyway to make this a setting too? (it could default to laion, but i imagine other research groups out there may want to try it for their own dataset / experiment name)

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+      config={
+      "learning_rate": args.learning_rate,
+      "architecture": "DiffusionPrior",
+      "dataset": "LAION-5B",
+      "epochs": 10,
+      })
+    print("wandb logging setup done!")
+       # Obtain the utilized device.
+    if torch.cuda.is_available():
```
lucidrains, line 139: anyway to do this condition only once?

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+    diffusion_prior = DiffusionPrior( net = prior_network, clip = None, image_embed_dim = image_embed_dim, 
+                                     timesteps = 100, cond_drop_prob = 0.2, loss_type = loss_type,
+                                     condition_on_text_encodings = condition_on_text_encodings).to(device)
+    # Gtext_reader image and text embeddings from the servers
+    print("==============Downloading embeddings - image and text====================")
+    image_reader = EmbeddingReader(embeddings_folder=image_embed_url, file_format="npy")
+    text_reader = EmbeddingReader(embeddings_folder=text_embed_url, file_format="npy")
+
+    ### Training code ###
+    optimizer = torch.optim.SGD(diffusion_prior.parameters(), lr = learning_rate)
```
lucidrains, line 38: `Adam` is probably better, and given we have a transformer here in the diffusion prior, you'd want to use `AdamW` with some specialized logic to omit the biases

lucidrains (commented):
looks like a good start! left some comments

dalle2_pytorch/train_diffusion_prior.py
lucidrains: https://github.com/lucidrains/tf-bind-transformer/blob/main/tf_bind_transformer/optimizer.py

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+            print("Training mse = ",loss.item())
+            train_loss+=loss.item()
+
+        print("Validation loop - epoch number ",e)
+        with torch.no_grad():
+            valid_loss = 0.0
+            val_set_size = int(val_percent*image_reader.count)
+            start = train_set_size
+            end=start+val_set_size
+            for emb_images,emb_text in zip(image_reader(batch_size=batch_size, start=start, end=end),
```
lucidrains, line 68: is this a separate set that isn't seen during training? (for your validation loop)

rom1504:
https://wandb.ai/laion/diffusion-prior/runs/gumsmjid?workspace=user-krish240574 training run

5300 steps in 35 minutes at batch size 10000 is 25 sample/s
That seems too fast to me.
any opinion on that @lucidrains ?

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
+import argparse
+import os
+from dalle2_pytorch import DiffusionPrior
+from embedding_reader import EmbeddingReader
```
rom1504, line 4: can you add that dependency in there https://github.com/lucidrains/DALLE2-pytorch/blob/main/setup.py#L24 ?

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+          batch_size,
+          train_percent,
+          val_percent,test_percent,
+          num_epochs,
+          loss_type,
+          condition_on_text_encodings,
+          device,
+          learning_rate=0.01):
+    # DiffusionPriorNetwork 
+    prior_network = DiffusionPriorNetwork( dim = image_embed_dim, depth = 6, dim_head = 64, heads = 8).to(device)
```
rom1504, line 26: adding these model params to the function params (and then the cli) would be best, that way they would be easy to tweak and they would be reported to wandb

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+    
+    # DiffusionPrior with text embeddings and image embeddings pre-computed
+    diffusion_prior = DiffusionPrior( net = prior_network, clip = None, image_embed_dim = image_embed_dim, 
+                                     timesteps = 100, cond_drop_prob = 0.2, loss_type = loss_type,
```
rom1504, line 30: same thing about these params

dalle2_pytorch/train_diffusion_prior.py
rom1504: I understand this is controlled by the start and end parameters here
The way embedding reader work is you give it a bunch of npy files (that can be located locally, at s3, at http or whatever else filesystem fsspec support), a start an end and a batch size and then it gives you batches

the start and end can indeed be used to split the dataset

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+
+                # Log to wandb
+                wandb.log({"Validation Loss(mse) ": loss})
+                valid_loss+=loss.item()
+
+        # Saving State Dict
+        torch.save(diffusion_prior.state_dict(), 'saved_model.pth')
+
+    test_set_size = int(test_percent*train_set_size)
+    with torch.no_grad():
```
rom1504, line 86: I would recommend to put that code in a function that can be reused for both validation and test (with different start and end settings)

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+            wandb.log({"Test mse ": loss})
+
+def main():
+    parser = argparse.ArgumentParser()
+    # Logging
+    parser.add_argument("--wandb-entity", type=str, default="laion")
+    parser.add_argument("--wandb-project", type=str, default="diffusion-prior")
+    # URLs for embeddings 
+    parser.add_argument("--image-embed-url", type=str, default="https://mystic.the-eye.eu/public/AI/cah/laion5b/embeddings/laion2B-en/img_emb/")
+    parser.add_argument("--text-embed-url", type=str, default="s3://laion-us-east-1/embeddings/vit-l-14/laion2B-en/text_emb/")
```
rom1504, line 108: can you point to https://mystic.the-eye.eu/public/AI/cah/laion5b/embeddings/laion2B-en/text_emb/ instead so that everyone has access ?

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+    # Hyperparamtext_readerers
+    parser.add_argument("--learning-rate", type=float, default=0.01)
+    parser.add_argument("--batch-size", type=int, default=10**6)
```
rom1504, line 111: oh are you really using 10**6 as a batch size ? I thought it was 10**4

rom1504:
I left a few comments as well, but looks good overall!

dalle2_pytorch/train_diffusion_prior.py
krish240574: Fixing this right away.

dalle2_pytorch/train_diffusion_prior.py
krish240574: It is testing at 10**4, that's correct. I put in 10**6 for the final run, I could default it to 10**4.

dalle2_pytorch/train_diffusion_prior.py
lucidrains: ohh got it, perfect :100: !

lucidrains:
> https://wandb.ai/laion/diffusion-prior/runs/gumsmjid?workspace=user-krish240574 training run
> 
> 5300 steps in 35 minutes at batch size 10000 is 25 sample/s That seems too fast to me. any opinion on that @lucidrains ?

it is reasonable, because the default settings lead to a rather small transformer `DiffusionPriorNetwork( dim = image_embed_dim, depth = 6, dim_head = 64, heads = 8)` only 6 layers - we should definitely scale this to 12 layers at least with 16 heads and 128 head dimension, once we validate that this prior network brings about the desired effect (more variety in the generations)

dalle2_pytorch/train_diffusion_prior.py
krish240574: Fixed.

dalle2_pytorch/train_diffusion_prior.py
krish240574: Fixed.

dalle2_pytorch/train_diffusion_prior.py
krish240574: Fixed. Now only checking once in main()

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,160 @@
...
+    args = parser.parse_args()
+    print("Setting up wandb logging... Please wait...")
+    wandb.init(
+      entity=args.wandb_entity,
+      project=args.wandb_project,
+      name=f"laion-dprior",
+      config={
+      "learning_rate": args.learning_rate,
+      "architecture": "DiffusionPrior",
+      "dataset": "LAION-5B",
```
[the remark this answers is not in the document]
krish240574: Fixed.

dalle2_pytorch/train_diffusion_prior.py
lucidrains: https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/optimizer.py#L8 just use this function and you are good to go!

dalle2_pytorch/train_diffusion_prior.py
krish240574: defaulting to 10 ** 4 now.

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,157 @@
...
+    min_valid_loss = np.inf
+    for e in range(epochs):
+        train_loss = 0.0 
+        print("Training loop - epoch number ",e)
+        train_set_size = int(train_percent*image_reader.count)
+        for emb_images,emb_text in zip(image_reader(batch_size=batch_size, start=0, end=train_set_size),
+                text_reader(batch_size=batch_size, start=0, end=train_set_size)):
+            emb_images = list(emb_images)
+            emb_text = list(emb_text)
+            print(emb_images[0].shape,emb_text[0].shape)
```
lucidrains, line 49: stray print?

krish240574:
All changes implemented - looks like we're good to go. I'll be glad to fix any more issues that may come up.

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,207 @@
...
+            condition_on_text_encodings = dp_condition_on_text_encodings).to(device)
+
+    # Get image and text embeddings from the servers
+    print("==============Downloading embeddings - image and text====================")
+    image_reader = EmbeddingReader(embeddings_folder=image_embed_url, file_format="npy")
+    text_reader = EmbeddingReader(embeddings_folder=text_embed_url, file_format="npy")
+    num_data_points = text_reader.count
+
+    ### Training code ###
+    optimizer = get_optimizer(diffusion_prior.parameters())
```
lucidrains, line 98: If you pull from the latest, you can just import from dalle2_pytorch.optimizers

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,207 @@
...
+                text_reader(batch_size=batch_size, start=start, end=end)):
+            emb_images = list(emb_images)
+            emb_text = list(emb_text)
+            emb_images[0] = torch.tensor(emb_images[0]).to(device)
+            emb_text[0] = torch.tensor(emb_text[0]).to(device)
+            model.eval()
+            loss = model(text_embed = emb_text[0],image_embed = emb_images[0])
+
+            # Log to wandb
+            if(isValidation):
```
lucidrains, line 46: Nit: you can just make the key in the dictionary a variable

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,207 @@
...
+
+    return AdamW(param_groups, lr = lr, weight_decay = wd)
+
+def eval_model(model,device,image_reader,text_reader,start,end,batch_size,isValidation=True):
+    with torch.no_grad():
+        for emb_images,emb_text in zip(image_reader(batch_size=batch_size, start=start, end=end),
+                text_reader(batch_size=batch_size, start=start, end=end)):
+            emb_images = list(emb_images)
+            emb_text = list(emb_text)
+            emb_images[0] = torch.tensor(emb_images[0]).to(device)
```
lucidrains, line 40: Why not just set a new variable, emb_image_tensor

dalle2_pytorch/train_diffusion_prior.py
lucidrains: @krish240574 last nitpick :)

lucidrains:
@krish240574 looking better! last round of comments and it will be merged

Thank you for this!

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,207 @@
...
+    parser.add_argument("--dp-cond-drop-prob", type=float, default=0.2)
+    parser.add_argument("--dp-loss-type", type=str, default="l2")
+    parser.add_argument("--clip", type=str, default=None)
+
+    args = parser.parse_args()
+    print("Setting up wandb logging... Please wait...")
+    wandb.init(
+      entity=args.wandb_entity,
+      project=args.wandb_project,
+      name=args.wandb_name,
```
rom1504, line 168: I recommend to remove this parameter
Otherwise all your runs end up being names the same
Without it, it will use random names which have the benefit of being unique

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,207 @@
...
+                text_reader(batch_size=batch_size, start=0, end=train_set_size)):
+            emb_images = list(emb_images)
+            emb_text = list(emb_text)
+            emb_images[0] = torch.tensor(emb_images[0]).to(device)
+            emb_text[0] = torch.tensor(emb_text[0]).to(device)
+            optimizer.zero_grad()
+            loss = diffusion_prior(text_embed = emb_text[0],image_embed = emb_images[0])
+            loss.backward()
+            # Log to wandb
+            wandb.log({"Training (mse)": loss})
```
rom1504, line 113: Would be great to report here as well:
* Number of steps (just add a step = 0 at the top and you do step= step+1 in the inner loop)
* Sample/s (you simply need a t = time.time() variable at the top and do sample_per_sec = step / (time.time() - t)

rom1504:
Looks good. I left 2 comments to improve a bit the wandb report.
I believe https://wandb.ai/laion/diffusion-prior/runs/2vfq744l?workspace=user-rom1504 is your latest run.
What I see:
* It's fast
* There are some spikes in training loss
* Training/test/val losses seem to go at 8*10^-4 of mse. Seems small, but we'll need more metrics and evaluation to know more #29 , that'll be for another PR

rom1504:
Ah wait there's one big thing that is missing: you need to save the checkpoints.
I recommend to save once every 30min and once at the end. You'll find it's much more convenient to save at a period of time than a number of step, so you don't have too few or too many checkpoints.

lucidrains:
@rom1504 yea, there's still a bunch of things missing actually, like gradient accumulation, gradient clipping, but its ok, i can take care of it

dalle2_pytorch/train_diffusion_prior.py
```diff
@@ -0,0 +1,207 @@
...
+            emb_text[0] = torch.tensor(emb_text[0]).to(device)
+            model.eval()
+            loss = model(text_embed = emb_text[0],image_embed = emb_images[0])
+
+            # Log to wandb
+            if(isValidation):
+                wandb.log({"Validation mse": loss})
+            else:
+                wandb.log({"Test mse": loss})
+
```
lucidrains, line 50: the loss type may not always be `mse`, and you can get the loss type identifier by `diffusion_prior.loss_type`

lucidrains:
@krish240574 so for training transformers, typically we also add gradient clipping for further stability https://github.com/lucidrains/x-transformers/blob/main/examples/enwik8_simple/train.py#L88 if you have no idea what that is, that's ok, i can just add it after this gets merged :pray:

krish240574:
> 

I'll look into it, I have a fair idea about gradient clipping.  I'll add it.

krish240574:
> Ah wait there's one big thing that is missing: you need to save the checkpoints. I recommend to save once every 30min (probably make 30 a param) and once at the end. You'll find it's much more convenient to save at a period of time than a number of step, so you don't have too few or too many checkpoints.

ok, I'll fix that.

lucidrains:
> > 
> 
> I'll look into it, I have a fair idea about gradient clipping. I'll add it.

thank you!

krish240574:
> @krish240574 looking better! last round of comments and it will be merged
> 
> Thank you for this!

Good to know, thanks for all the awesome code !

lucidrains:
> > @krish240574 looking better! last round of comments and it will be merged
> > Thank you for this!
> 
> Good to know, thanks for all the awesome code !

thank you for contributing :pray:

krish240574:
> > > 
> > 
> > 
> > I'll look into it, I have a fair idea about gradient clipping. I'll add it.
> 
> thank you!

my pleasure!

krish240574:
> @krish240574 so for training transformers, typically we also add gradient clipping for further stability https://github.com/lucidrains/x-transformers/blob/main/examples/enwik8_simple/train.py#L88 if you have no idea what that is, that's ok, i can just add it after this gets merged pray

Hey @lucidrains, I'm thinking I should put this inside the main training loop, implemented as in the x-transformers code - but to reach the CausalTransformer inside the DiffusionPriorNetwork - 
would that then be                 
torch.nn.utils.clip_grad_norm_(diffusion_prior.net.causal_transformer.parameters(), 0.5)  ?

- I'm not a 100% sure if that is right, do correct me if I'm wrong. Would you be open to adding it yourself, I don't want to break anything. I apologize for the inconvenience.

lucidrains:
> > @krish240574 so for training transformers, typically we also add gradient clipping for further stability https://github.com/lucidrains/x-transformers/blob/main/examples/enwik8_simple/train.py#L88 if you have no idea what that is, that's ok, i can just add it after this gets merged pray
> 
> @lucidrains, I'm thinking I should put this inside the main training loop, implemented as in the x-transformers code - but to reach the CausalTransformer inside the DiffusionPriorNetwork - would that then be torch.nn.utils.clip_grad_norm_(diffusion_prior.net.causal_transformer.parameters(), 0.5) ?
> 
> * I'm not a 100% sure if that is right, do correct me if I'm wrong. Would you be open to adding it yourself, I don't want to break anything. I apologize for the inconvenience.

ahh no problem i'll take care of it

if @rom1504 says it is good, we'll merge it!

rom1504:
LGTM now

@lucidrains I advise you choose the squash mode when merging this one to have a single commit in master

rom1504:
<https://wandb.ai/laion/diffusion-prior/runs/3ik8aq8l?workspace=user-krish240574> latest run

lucidrains:
thank you both! 🙏

rom1504:
<https://github.com/lucidrains/DALLE2-pytorch/commits/main> oh well I guess squashing will be for another pr :D

lucidrains:
NOOO! you just told me loll

sorry, i messed up 😢 😆

lucidrains:
how do i fix this? lemme see

lucidrains:
@krish240574 do you want to see if you could redo a PR from your branch? i did a hard reset, my bad

rom1504:
lol sorry I should have been clearer

Screenshot_20220501_204118

here is what I meant

lucidrains:
> lol sorry I should have been clearer
> 
> Screenshot_20220501_204118
> 
> here is what I meant

yea i know, just misclick lol

PR #53, opened 2022-05-03
Missing imports and nn.utils.clip_grad_norm_()

krish240574:
The following imports are needed :
from torch.cuda.amp import autocast,GradScaler

nn.utils.clip_grad_norm_ works, nn.init.clip_grad_norm gives an AttributeError.


lucidrains:
Oops, yes you are right!

PR #55, opened 2022-05-03
Reporting metrics - Cosine similarity. 

krish240574:
More info being logged into wandb - cosineSim(text, pred_image), cosineSim(text, orig_image), 
cosineSim(text, pred_image) - cosineSim(text, orig_image).


train_diffusion_prior.py
```diff
@@ -137,6 +163,17 @@ def train(image_embed_dim,
...
+            # Use NUM_TEST_EMBEDDINGS samples from the test set each time
+            # Get embeddings from the most recently saved model
+            diff_cosine_sim = report_cosine_sims(diffusion_prior,
+                    image_reader,
+                    text_reader,
+                    train_set_size,
+                    val_set_size,
+                    NUM_TEST_EMBEDDINGS,
+                    device)
+            wandb.log({"Cosine similarity difference": diff_cosine_sim})
```
rom1504, line 176: Let's run this once every N steps (default N = 100 )
And let's do that as well for the computation of the validation loss instead of only at the end like now

rom1504:
I think we should have something a bit more generic where we compute several metrics for both the eval and test set every N=100 steps.
No reason to do something different for eval and test about metric computation
The only difference is that test metrics shouldn't be used to select HP params

krish240574:
Please change lines 64-68 to the following  : 
        original_similarity = cos(text_embed,test_image_embeddings).cpu().numpy()
        predicted_similarity = cos(text_embed,predicted_image_embeddings).cpu().numpy()

        wandb.log({"CosineSimilarity(text_embed,image_embed)": np.mean(original_similarity)})
        wandb.log({"CosineSimilarity(text_embed,predicted_image_embed)":np.mean(predicted_similarity)})

I realised I had submitted the PR before making those changes. My apologies.

rom1504:
@krish240574 you can push further changes to this branch

krish240574:
@rom1504, just realised that - brain freeze ! Pushed changes.

lucidrains:
@krish240574 @rom1504 thank you Kumar, and Romain for the code review!

PR #57, opened 2022-05-04
Add a dataloader for training the decoder

Veldrovive:
This dataloader loads images and embeddings from a webdataset by extending the standard WebDataset pipeline with a stage to load embeddings from npy files. Since embeddings are stored separately from the images in the case of LAION 5B, we also include the capability to read embeddings from a different source as long as the shards in the webdataset align with the shards in the embeddings.


lucidrains:
boss. :100: :1st_place_medal:

lucidrains:
@Veldrovive i'll let Romain take a look at it and let's aim to merge it this week!

setup.py
```diff
@@ -33,9 +33,11 @@
     'torchvision',
     'tqdm',
     'vector-quantize-pytorch',
-    'webdataset',
-    'x-clip>=0.5.1',
-    'youtokentome'
+    'x-clip>=0.4.4',
+    'youtokentome',
+    'webdataset>=0.2.5',
+    'requests>=2.23.0',
```
rom1504, line 39: I don't think we need requests, do we ?

Veldrovive: You're right. I will remove that.

dalle2_pytorch/dataloaders/decoder_loader.py
```diff
@@ -0,0 +1,208 @@
...
+    file_paths_with_prefix = [prefix + file_path for file_path in file_paths]
+    return fs, file_paths_with_prefix
+### Code from embedding_reader ends
+
+def embedding_inserter(samples, embeddings_url, shard_width, handler=wds.handlers.reraise_exception):
+    """Given a datum of {"__key__": str, "__url__": str, ...} adds the cooresponding embedding and yields"""
+    previous_tar_url = None
+    current_embeddings = None
+    # Get a reference to an abstract file system where the embeddings are stored
+    fs, embedding_files = get_file_list(embeddings_url, 'npy')
```
rom1504, line 66: Do we need the full list here ? It seems you're only using the first item

Veldrovive: Ah, that's relic of using a different method of finding the corresponding embedding file. If I change this to a simpler method of getting the paths, it would also mean I could remove the heavy functions above and simplify the thing a bit at the cost of a little bit of safety.

dalle2_pytorch/dataloaders/decoder_loader.py
```diff
@@ -0,0 +1,208 @@
...
+            assert "jpg" in sample, f"Sample {sample['__key__']} missing image"
+            assert "npy" in sample, f"Sample {sample['__key__']} missing embedding. Did you set embedding_folder_url?"
+            yield sample
+        except Exception as exn:  # From wds implementation
+            if handler(exn):
+                continue
+            else:
+                break
+
+class ImageEmbedingDataset(wds.DataPipeline, wds.compat.FluidInterface):
```
lucidrains, line 117: small typo here `Embeding` :smile:

Veldrovive: Gah, thought I caught all of those. I guess it wouldn't be me if there weren't a couple left. Fix incoming.

dalle2_pytorch/dataloaders/decoder_loader.py
Veldrovive: Check the newest commit for this change.

rom1504:
now lgtm

however one more thing, @Veldrovive could you add a note in a new section in the readme as to
1. how to use that data loader (initialization + a loop)
2. a short description of the data (webdataset + npy files) and how to produce it (eg img2dataset + clip retrieval to produce embeddings)
so we don't forget how to use this

Veldrovive:
I added an example. I also added an `__init__.py` for the data loaders, but this was a stylistic choice so I can remove that again if we want the data loaders to be imported with their full path.

rom1504:
Nice!
I think it's ready for merging

Next step (in other PRs) will be to make a simple (non distributed) training script :)

lucidrains:
@Veldrovive @rom1504 thank you Aidan and Romain!

PR #66, opened 2022-05-06
simulate unrelated captions as a training metric

nousr:
shuffle the text embeddings along the first axis, predict on shuffled embedding, then compare your predicted image embedding to the original text embedding to simulate an unrelated caption/pred_img

this trick assumes the distribution of embeddings is uniform, but given we're working on the billions scale I think that's alright 😄


lucidrains:
@nousr cool! do you think you can just get with a shift? that would guarantee a mismatch of embeddings

```python
import torch
batch_size = 1000
torch.roll(torch.arange(batch_size), 1)
```

nousr:
@lucidrains thanks for the suggestion :D should be good now 👍

lucidrains:
Thank you!

PR #67, opened 2022-05-06
Val_loss every N steps, model loading from pre-trained, hyperpmts reporting. 

krish240574:
In train_diffusion_prior.py
- Validation loss is getting every N steps, for a small batch of 100 embeddings
- Model loading implemented
- hyperparameters logging to wandb for each run
- Some code cleanup in the report_cosine_sims() function. 

I can't figure out, why the addition I made to README.md shows up again here.


train_diffusion_prior.py
```diff
@@ -125,6 +127,13 @@ def train(image_embed_dim,
     # Create save_path if it doesn't exist
     if not os.path.exists(save_path):
         os.makedirs(save_path)
+        
+    # Load pre-trained model from DPRIOR_PATH
+    if RESUME:
```
rom1504, line 132: That's assuming the default model parameters are used

It would be better to :
* Save the hp parameters (all the kwargs of diffusionprior) into:
  * Wandb
  * The checkpoint (as another dict besides state dict)

That way we would know the params both from wandb and from the checkpoint

rom1504:
I advise in general you use another branch that main to avoid merging issue (make one new branch for each new feature)

The metric change lgtm

For the hp param saving more is needed, left an inline comment

train_diffusion_prior.py
krish240574: ok, gotcha. Let me fix that.

krish240574: Fixed. Please take a look.

train_diffusion_prior.py
```diff
@@ -252,20 +253,31 @@ def main():
...
+    config = ({"learning_rate": args.learning_rate,
       "architecture": args.wandb_arch,
       "dataset": args.wandb_dataset,
       "Learning Rate":args.learning_rate,
       "Weight Decay":args.weight_decay,
       "Max Gradient Clipping Norm":args.max_grad_norm,
       "Batch size":args.batch_size,
-      "epochs": args.num_epochs
-      })
+      "epochs": args.num_epochs,
```
rom1504, line 264: can you use these params in the resume part ?

krish240574: Ok, fixed - the hyperparameters of the pretrained model are now being used - only in case of RESUME.

nousr:
can we store the current position of the embedding reader when saving the model. That way it's possible to resume training and be able to pick up where you left off so to speak.

train_diffusion_prior.py
rom1504: No you didn't understand what I mean.

In resume you are not using the hyperparameters from the checkpoint
You are still using the parameters from argparse
https://github.com/lucidrains/DALLE2-pytorch/blob/6ef5872a35ff01c55da80c6cf9b3c4ee5ffb31c0/train_diffusion_prior.py#L107 there

When doing that, you will notice it would be much more convenient to have a subdict for each network with exactly the same names as the constructor of the network so let's do that as well

krish240574: Fixed, do take a look .

train_diffusion_prior.py
```diff
@@ -237,26 +298,44 @@ def main():
...
-    wandb.init(
-      entity=args.wandb_entity,
-      project=args.wandb_project,
-      config={
-      "learning_rate": args.learning_rate,
+    
+    config = ({"learning_rate": args.learning_rate,
       "architecture": args.wandb_arch,
       "dataset": args.wandb_dataset,
+      "Learning Rate":args.learning_rate,
```
rom1504, line 305: duplicated

train_diffusion_prior.py
```diff
@@ -237,26 +298,44 @@ def main():
...
+      "Weight Decay":args.weight_decay,
+      "Max Gradient Clipping Norm":args.max_grad_norm,
+      "Batch size":args.batch_size,
       "epochs": args.num_epochs,
+      "Diffusion_Prior_Network":{"dpn_depth":args.dpn_depth,
+      "dpn_dim_head":args.dpn_dim_head,
+      "dpn_heads":args.dpn_heads},
+      "Diffusion_Prior":{"dp_condition_on_text_encodings": args.dp_condition_on_text_encodings,
```
rom1504, line 313: let's use the same names as the constructor

train_diffusion_prior.py
```diff
@@ -41,16 +41,64 @@ def eval_model(model,device,image_reader,text_reader,start,end,batch_size,loss_t
         avg_loss = (total_loss / total_samples)
         wandb.log({f'{phase} {loss_type}': avg_loss})
 
-def save_model(save_path, state_dict):
+def save_diffusion_model(save_path, state_dict):
     # Saving State Dict
     print("====================================== Saving checkpoint ======================================")
     torch.save(state_dict, save_path+'/'+str(time.time())+'_saved_model.pth')
+    
+def load_diffusion_model(dprior_path,
```
rom1504, line 49: let's put that in dalle2_pytorch.py so it can be reused

train_diffusion_prior.py
```diff
@@ -156,9 +214,9 @@ def train(image_embed_dim,
             if(int(time.time()-t) >= 60*save_interval):
                 t = time.time()
 
-                save_model(
+                save_diffusion_model(
                     save_path,
-                    dict(model=diffusion_prior.state_dict(), optimizer=optimizer.state_dict(), scaler=scaler.state_dict()))
+                    dict(model=diffusion_prior.state_dict(), optimizer=optimizer.state_dict(), scaler=scaler.state_dict(),hyperpmts = config))
```
rom1504, line 219: let's put  that in the function

rom1504:
@nousr do you feel like finishing up this PR?
krish will be a bit busy in the next days

train_diffusion_prior.py
krish240574: Fixed.

train_diffusion_prior.py
krish240574: Fixed.

train_diffusion_prior.py
krish240574: Fixed.

dalle2_pytorch/dalle2_pytorch.py
```diff
@@ -72,6 +72,50 @@ def pad_tuple_to_length(t, length, fillvalue = None):
         return t
     return (*t, *((fillvalue,) * remain_length))
 
+# Diffusion prior model loading and saving helper functions
+
+def load_diffusion_model(dprior_path, image_embed_dim, device, wandb_entity, wandb_project):
```
rom1504, line 77: can you remove the wandb params and move image_embed_dim param in the checkpoint ?

train_diffusion_prior.py
rom1504: looks like you changed the function, but the call still need to be adapted

train_diffusion_prior.py
krish240574: Quick point - having the same names as the constructor might not work - to do a **dpn_config , because the DPN constructor uses a parameter from the dp_config(i.e  dp_normformer).

train_diffusion_prior.py
krish240574: aargh, let me fix that

krish240574: Done.

dalle2_pytorch/dalle2_pytorch.py
krish240574: Ok, will do .

train_diffusion_prior.py
krish240574: fixed

PR #79, opened 2022-05-09
Ensure Eval Mode In Metric Functions

nousr:
Loss started looking like 📈 and @rom1504 noticed we weren't properly setting eval/train mode

### DIFF
--- a/train_diffusion_prior.py
+++ b/train_diffusion_prior.py
@@ -41,10 +41,10 @@ def eval_model(model,device,image_reader,text_reader,start,end,batch_size,loss_t
         avg_loss = (total_loss / total_samples)
         wandb.log({f'{phase} {loss_type}': avg_loss})
 
-	model.train()
+    model.train()
 
 def report_cosine_sims(diffusion_prior,image_reader,text_reader,train_set_size,NUM_TEST_EMBEDDINGS,device):
-	diffusion_prior.eval()
+    diffusion_prior.eval()
 
     cos = nn.CosineSimilarity(dim=1, eps=1e-6)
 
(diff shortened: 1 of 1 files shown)


lucidrains:
@nousr ohh i see, instead of adding these, i would move https://github.com/lucidrains/DALLE2-pytorch/blob/79f388d106c0f8d9ad3b3bfe5a484b9b84a576f2/train_diffusion_prior.py#L175 to line 179 instead 🙏

lucidrains:
it looks like dropout isn't being used in the training script, so the training vs eval mode wouldn't matter

lucidrains:
addressed the lack of dropout issue https://github.com/lucidrains/DALLE2-pytorch/commit/a774bfefe27ceb6171ad486165739014cf1ee9f0 so dropout is beginning to look less and less useful for large transformers, but may still have a place if one is training sub 1B parameter models

nousr:
@lucidrains gotcha gotcha

I shifted around the train toggle, and kept the eval call I added inside of ```get_cosine_similarity()```.

this was prompted by the charts in krish's lastest run https://wandb.ai/laion/diffusion-prior/runs/1blxu24j?workspace=user-nousr_laion

not sure what happened there, especially considering multi-million point runs have been completed before without this crazy jump in loss

lucidrains:
@nousr this happens all the time with transformers training :) not that unusual

some groups even roll back before the divergence and retry at a lower learning rate 😆

lucidrains:
@nousr looking good! merging, thank you!

lucidrains:
@nousr if you are going to redo the experiment from scratch, do recommend turning on the `normformer` setting, which purportedly stabilizes transformer training for some practitioners

PR #84, opened 2022-05-11
Decoder basic training script

Veldrovive:
This is a draft PR to ensure that ensure the general idea of decoder training is being done correctly before hammering in the details and adding CLI arguments.

There is still test code that overfits the model and verifies the embeddings are being read correctly. 
Example training is available here: https://wandb.ai/veldrovive/dalle2_train_decoder
It is obvious that overfitting works, but the first more significant test run with 1 million image-embedding pairs is still a work in progress.

Most issues may appear in the `create_decoder` or `train` functions in the train_decoder.py file.


lucidrains:
@Veldrovive thank you for getting started with this Aidan! :pray: i'm also investigating the learned variance approach in gaussian diffusion class, which i'm not using yet. @borisdayma alerted me to this, and it seems Glide may be using it, so i'll bring it in once initial results locally show an improvement :)

README.md
```diff
@@ -587,6 +587,47 @@ images = dalle2(
...
 
+## Dataloaders
+In order to make loading data simple and efficient, we include some general dataloaders that can be used to train portions of the network.
+
+### Decoder: Image Embedding Dataset
+When training the decoder (and up samplers if training together) in isolation, you will need to load images and corresponding image embeddings. This dataset can read two similar types of datasets. First, it can read a [webdataset](https://github.com/webdataset/webdataset) that contains `.jpg` and `.npy` files in the `.tar`s that contain the images and associated image embeddings respectively. Alternatively, you can also specify a source for the embeddings outside of the webdataset. In this case, the path to the embeddings should contain `.npy` files with the same shard numbers as the webdataset and there should be a correspondence between the filename of the `.jpg` and the index of the embedding in the `.npy`. So, for example, `0001.tar` from the webdataset with image `00010509.jpg` (the first 4 digits are the shard number and the last 4 are the index) in it should be paralleled by a `img_emb_0001.npy` which contains a NumPy array with the embedding at index 509.
```
srelbo, line 594: Thanks for adding the `webdataset` support! 👍

train_decoder.py
```diff
@@ -0,0 +1,488 @@
...
+        index_width=4,
+        train_prop = 0.75,
+        val_prop = 0.15,
+        test_prop = 0.10
+    )
+
+    decoder, (unet_config, decoder_config) = create_decoder(device)
+
+    wandb.init(
+        entity="Veldrovive",
```
srelbo, line 382: ^ Perhaps get this from environment or somewhere else, so others can use your trainer?

dalle2_pytorch/dataloaders/decoder_loader.py
```diff
@@ -112,28 +157,43 @@ def __init__(
...
+        if embedding_folder_url is not None:
+            # There may be webdataset shards that do not have a embedding shard associated with it. If we do not skip these, they would cause issues.
+            self.append(skip_unassociated_shards(embeddings_url=embedding_folder_url, handler=handler))
 
         self.append(wds.split_by_node)
         self.append(wds.split_by_worker)
 
         self.append(wds.tarfile_to_samples(handler=handler))
-        self.append(wds.decode("torchrgb"))
+        self.append(wds.decode("pilrgb", handler=handler))
```
srelbo, line 169: Just curious, why this change ^ ?

train_decoder.py
Veldrovive: Yes, this is just to get some feedback on if the training is set up correctly. CLI args are the next step once it produces any results on a larger dataset.

dalle2_pytorch/dataloaders/decoder_loader.py
srelbo: Makes sense, thank you for the explanation @Veldrovive !

Veldrovive:
I updated the training script to take parameters. Since the config for the model gets very complicated and cannot be expressed in a few parameters, I opted to define a config file instead. I see this is not usually the style @lucidrains has been using for these replications so if it should be converted to CLI arguments only I can make a less configurable version that meets the requirements.

If config files are fine, I would suggest we move the training scripts into a dedicated directory so that they don't bloat the root directory.

train_decoder.py
Veldrovive: The option to use any entity has now been added along with other configuration options.

Veldrovive:
I have also noticed that loading the `state_dict` of the trainer does not seem to load the `state_dict`s of the optimizers and scalers as there is always a jump in loss after loading an old model. I will try loading them directly into the relevant objects instead of to the trainer.

lucidrains:
> @Veldrovive thank you for getting started with this Aidan! 🙏 i'm also investigating the [learned variance approach](https://github.com/hojonathanho/diffusion/blob/master/diffusion_tf/diffusion_utils_2.py#L134) in gaussian diffusion class, which i'm not using yet. @borisdayma alerted me to this, and it seems Glide may be using it, so i'll bring it in once initial results locally show an improvement :)

this is now done :)

rom1504:
this pr looks ok enough

anything more you want to do before merging @Veldrovive ?

Veldrovive:
I just want to hammer out the unexpected behavior of loading the trainer not loading the optimizer state and make sure @lucidrains is ok with having configuration passed via config files instead of command line arguments. After that, it will be ready for review.

lucidrains:
@Veldrovive yea sure! maybe we can use @rom1504 's https://github.com/criteo/fromconfig

lucidrains:
@Veldrovive so the only thing is the reliance on `wandb`, i'd rather have the logic modularized this time around https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/trackers.py but its ok if you think it is too much work (i can always refactor it later)

train_decoder_config.json.example
```diff
@@ -0,0 +1,86 @@
+{
```
lucidrains, line 1: maybe move this to a `configs/` folder

train_decoder.py
```diff
@@ -0,0 +1,462 @@
...
+        resume_config=config["resume"],
+        using_wandb=using_wandb,
+        **config["train"],
+    )
+
+
+class TrainDecoderConfig:
+    def __init__(self, config):
+        defaults = {
+            "unets": [{ "dim": 16, "image_emb_dim": 768, "cond_dim": 64, "channels": 3, "dim_mults": [1, 2, 3, 4], "attn_dim_head": 32, "attn_heads": 16 }],
```
lucidrains, line 386: hmm, couldn't this also be moved to a config?

train_decoder.py
```diff
@@ -0,0 +1,462 @@
...
+    with torch.no_grad():
+        decoder.eval()
+        for data in test_iter:
+            if len(data) == 3:
+                img, emb, txt = data
+            else:
+                img, emb = data
+                txt = [""] * emb.shape[0]
+            # If the number of images is larger than n, we can remove all embeddings after the first n to speed up inference
+            if img.shape[0] > n:
```
lucidrains, line 105: why are we doing this instead of just having the iterator return it in batches of `n`

train_decoder.py
```diff
@@ -0,0 +1,462 @@
...
+            sample = decoder.sample(emb)
+            for original_image, generated_image, text in zip(img, sample, txt):
+                # Make a grid containing the original image and the generated image
+                img_grid = torchvision.utils.make_grid([original_image, generated_image])
+                image = wandb.Image(img_grid, caption=text_prepend+text)
+                images.append(image)
+                
+                if len(images) >= n:
+                    decoder.to(device)  # Again, don't ask me why but something here moves the model to the cpu
+                    decoder.train()
```
lucidrains, line 121: `train()` is best called just before the training step

train_decoder.py
```diff
@@ -0,0 +1,462 @@
...
+        last_time = time.time()
+        losses = []
+        for i, (img, emb) in enumerate(dataloaders["train"]):
+            step += 1
+            sample += img.shape[0]
+            img, emb = send_to_device((img, emb))
+            
+            for unet in range(1, trainer.num_unets+1):
+                loss = trainer.forward(img, image_embed=emb, unet_number=unet)
+                loss.backward()
```
lucidrains, line 242: this is no longer necessary in the newer version

train_decoder.py
```diff
@@ -0,0 +1,462 @@
...
+        with torch.no_grad():
+            decoder.eval()
+            sample = 0
+            average_loss = 0
+            start_time = time.time()
+            for i, (img, emb, txt) in enumerate(dataloaders["val"]):
+                sample += img.shape[0]
+                img, emb = send_to_device((img, emb))
+                
+                with autocast(enabled=amp):
```
lucidrains, line 290: the trainer class should take care of this

lucidrains:
> this pr looks ok enough
> 
> anything more you want to do before merging @Veldrovive ?

haha, i think we could polish this PR a bit more before merging actually

train_decoder.py
Veldrovive: Yes, the defaults could be a separate file instead of hard coded. I generally prefer them to be hard coded so that people don't change the defaults instead of creating their own config, but if you prefer them to be separated I can make a separate file for them.

train_decoder.py
```diff
@@ -0,0 +1,462 @@
...
+    test_iter = iter(dataloader)
+    images = []
+    with torch.no_grad():
+        decoder.eval()
+        for data in test_iter:
+            if len(data) == 3:
+                img, emb, txt = data
+            else:
+                img, emb = data
+                txt = [""] * emb.shape[0]
```
lucidrains, line 103: what is this doing exactly? is this to accommodate logic between text conditioned decoder vs without?

train_decoder.py
lucidrains: yea, i think if you want to go config driven, just go all the way, otherwise we keep them as click argument defaults

train_decoder.py
```diff
@@ -0,0 +1,462 @@
...
+        last_sample = 0
+        last_snapshot = 0
+        last_time = time.time()
+        losses = []
+        for i, (img, emb) in enumerate(dataloaders["train"]):
+            step += 1
+            sample += img.shape[0]
+            img, emb = send_to_device((img, emb))
+            
+            for unet in range(1, trainer.num_unets+1):
```
lucidrains, line 240: so the other thing is, different unets in the decoder may require different amounts of training, so the researcher may have to specify which specific unet they want trained in the config

it is one of the pros of the cascading ddpm (but perhaps also the training script will be slightly more involved, as one is training all different unets to completion)

train_decoder.py
Veldrovive: The dataloader allows other files to be loaded from the webdataset beyond the image and embedding. In this case, the test dataset loads caption files so that it can be logged with the sample. Since we also sample the training set at some points, the text is not always available and a default is used. The other option would be to always load the text file as well. It would not have much of an effect beyond making this part a one liner.

train_decoder.py
lucidrains: oh actually, nevermind, i don't handle this just yet, let me think about it

train_decoder.py
Veldrovive: Is isolating the unets as simply as only running the forward pass on the unets that are being trained? If that's the case I can put that in the config.

train_decoder.py
lucidrains: ok, now it should handle it, as long as you place the trainer into eval mode (then call forward)

train_decoder.py
lucidrains: ohh i see, is there anyway to detect this early on and just pass in a `has_text` boolean so it is more clear? (instead of relying on the length of the tuple returned)

train_decoder.py
Veldrovive: This is so that the existing training set can also be sampled from without always using the full batch size as that is very slow. A separate loader could be created for the same purpose I suppose.

train_decoder.py
lucidrains: @Veldrovive yup, you simply pass the appropriate `unet_number`  for `forward`, and then `update`

lucidrains:
otherwise looks good! let's aim to merge this mid-to-late-next week :)

train_decoder.py
lucidrains: @Veldrovive ohh yup, let's use a separate dataloader with the proper `sampling_batch_size`

Veldrovive:
> @Veldrovive so the only thing is the reliance on `wandb`, i'd rather have the logic modularized this time around https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/trackers.py but its ok if you think it is too much work (i can always refactor it later)

I'm always in favor of abstracting logging. I can use the trackers, but I am not sure of their scope in some cases. In particular, logging media and saving models. It would be nice to be able to do these through the tracker so that there is less code that relies on knowing if you are actually using wandb, but currently you would still need to do the check before converting media to the wandb image format or whether to call wandb.save on your model checkpoint.

train_decoder.py
```diff
@@ -0,0 +1,462 @@
...
+                    decoder.to(device)  # Again, don't ask me why but something here moves the model to the cpu
+                    decoder.train()
+                    return images
+    decoder.train()  # In case we run out of samples before n. Your dataset must be tiny, but whatever. Edge cases.
+
+def save_trainer(base_path, trainer, epoch, step, validation_losses, local_only=True, latest=False, best=False, checkpoint=False):
+    """
+    Saves the state of the trainer and decoder to wandb.
+    """
+    print("====================================== Saving trainer ======================================")
```
lucidrains, line 129: haha kind of nit, but can you use the `print_ribbon` function from https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/train.py#L100

lucidrains:
> I have also noticed that loading the `state_dict` of the trainer does not seem to load the `state_dict`s of the optimizers and scalers as there is always a jump in loss after loading an old model. I will try loading them directly into the relevant objects instead of to the trainer.

yea the other thing (and this will have to be done for the diffusion prior training) is saving the EMA models. for decoder, each of the unets have their corresponding EMA

lucidrains:
> > @Veldrovive so the only thing is the reliance on `wandb`, i'd rather have the logic modularized this time around https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/trackers.py but its ok if you think it is too much work (i can always refactor it later)
> 
> I'm always in favor of abstracting logging. I can use the trackers, but I am not sure of their scope in some cases. In particular, logging media and saving models. It would be nice to be able to do these through the tracker so that there is less code that relies on knowing if you are actually using wandb, but currently you would still need to do the check before converting media to the wandb image format or whether to call wandb.save on your model checkpoint.

yea, if its too complicated, i can handle it. i have a plan for the maybe wrapping with `wandb` types, as well as logic for whether and where to log images to disk

lucidrains:
> > > @Veldrovive so the only thing is the reliance on `wandb`, i'd rather have the logic modularized this time around https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/trackers.py but its ok if you think it is too much work (i can always refactor it later)
> > 
> > 
> > I'm always in favor of abstracting logging. I can use the trackers, but I am not sure of their scope in some cases. In particular, logging media and saving models. It would be nice to be able to do these through the tracker so that there is less code that relies on knowing if you are actually using wandb, but currently you would still need to do the check before converting media to the wandb image format or whether to call wandb.save on your model checkpoint.
> 
> yea, if its too complicated, i can handle it. i have a plan for the maybe wrapping with `wandb` types, as well as logic for whether and where to log images to disk

it'll be worth it, since in projects past, there was a difficult time with using weights and biases in a distributed environment

Veldrovive:
Ok. I don't want to step on any toes so for now I will just move logging into its own functions so that when the tracker system is complete it should not be difficult to transfer to using those methods instead.

lucidrains:
@Veldrovive yup, even if you just place all the wandb related stuff into the `WandbTracker`, forward the function calls, but noop the `ConsoleTracker`, that would be good!

lucidrains:
> I have also noticed that loading the `state_dict` of the trainer does not seem to load the `state_dict`s of the optimizers and scalers as there is always a jump in loss after loading an old model. I will try loading them directly into the relevant objects instead of to the trainer.

eventually, once a clean solution clicks in my mind, i'll just add `save` and `load` methods on the respective trainers (but i need to figure out how to handle the configurations of the unets, vqgan-vae (if they exist), and clip

train_decoder.py
```diff
@@ -0,0 +1,495 @@
...
+        else:
+            raise ValueError("resume_from must be either wandb or file")
+    trainer = trainer.to(device=inference_device)
+    
+    send_to_device = lambda arr: [x.to(device=inference_device, dtype=torch.float) for x in arr]
+    step = start_step
+    for epoch in range(start_epoch, epochs):
+        print(print_ribon(f"Starting epoch {epoch}", repeat=40))
+        trainer.train()
+        decoder.train()  # I don't see anything in the trainer that would set decoder to train mode so I assume we also need this
```
lucidrains, line 261: oh no, you won't need this, because the trainer is also a `nn.Module` class, so it will propagate the train signal to all of its children

train_decoder.py
```diff
@@ -0,0 +1,495 @@
...
+
+
+def create_decoder(device, decoder_config, unets_config):
+    """Creates a sample decoder"""
+    unets = []
+    for i in range(0, len(unets_config)):
+        unets.append(Unet(
+            **unets_config[i]
+        ))
+        unets[i].to(device=device)
```
lucidrains, line 79: i'm pretty sure when you do `decoder.to(device)` below, it will set all the unets too

train_decoder.py
Veldrovive: I need some documentation on how `to`, `train`/`eval`, and `state_dict` propagate. I assumed at first that they all propagate, but that was wrong because `state_dict` doesn't at least. Then an error resolved itself when I separately sent the trainer and decoder to the device, but I guess that was coincidental and `to` does actually propagate.

I'll make a note to go through and remove calls to anything that may be propagating.

train_decoder.py
```diff
@@ -0,0 +1,495 @@
+from dalle2_pytorch import Unet, Decoder
+from dalle2_pytorch.train import DecoderTrainer, print_ribon
```
lucidrains, line 2: `ribon` should be `ribbon`

lucidrains:
sample-37

the decoder trains for me, on the unconditional setting. here is some results at 16k for oxford flowers dataset - they look quite good!

recommend that you merge in the latest commit

also recommend that you simply save the `state_dict` of the entire decoder training object. it will contain the optimizer, the EMA models, all unets, as well as the training step number

configs/train_decoder_config.json.example
```diff
@@ -0,0 +1,87 @@
...
+                "size": [128, 128],
+                "scale": [0.75, 1.0],
+                "ratio": [1.0, 1.0]
+            },
+            "ToTensor": true
+        }
+    },
+    "train": {
+        "epochs": 100,
+        "lr": 2e-5,
```
lucidrains, line 58: in the paper, this was `1e-4`, i've corrected it in the latest

also found out that `3e-4` may be too high (had a diverged run)

xiankgx:
I can confirm the decoder seems to work alright now after the change back to ResNet blocks from ConvNext blocks.

lucidrains:
@Veldrovive ohh yes, you are right, optimizer may have to be separately saved and loaded

i'll have to get to the saving and loading methods on the trainer at a later date, if circumstances allow

Veldrovive:
@lucidrains When refactoring to extract tracker logic, I just want to clarify a couple scope questions. Is the purpose to abstract everything that may have to do with a tracker service like wandb (i.e. logging runtime data, media, saving models, and recalling models from the server) or is it supposed to be very specific to just logging runtime data?

lucidrains:
@Veldrovive yup, anything that has to do with experimental tracking. basically it would be nice for those who do not wish to use wandb to just instantiate the console tracker, and media is saved locally, losses printed to stdout, other operations nooped if appropriate

Veldrovive:
I converted the script to work with trackers. There's currently redundancy in some functions because this is a wip, but the main structure is there and I wanted to get it in so that people could take a look.

train_decoder.py
Veldrovive: I improved this by having the dataset expose which keys it is using.
https://github.com/Veldrovive/DALLE2-pytorch/blob/61ceb19045ce3d6e77cbaabd6968e720da2dbc8f/train_decoder.py#L106-L114

train_decoder.py
Veldrovive: The training configuration now has a `unet_training_mask` which is a boolean array that defines which unets will be trained.

train_decoder.py
Veldrovive: I updated this so that the sampling code uses separate data loaders with a config that ensures consistency.

train_decoder.py
Veldrovive: I removed the defaults to a `decoder_defaults.py` file. It's a python file instead of json because that allows for comments on the defaults, using more complex objects like a `REQUIRED` enum that makes it possible to have default values of None, and easier importing.

Veldrovive:
I think everything that was brought up has now been addressed. There is still a bug with loading the old optimizers/scalers/ema models so the only thing that is currently preserved is the state of the decoder itself.

Veldrovive:
FID, IS, KID, and LPIPS, the generative image metrics available from torchmetrics, have been implemented and you can see a wandb run with all enabled [here](https://wandb.ai/veldrovive/dalle2_train_decoder/runs/30hxfjrp?workspace=user-veldrovive). Currently, by default none are enabled, but it would also be reasonable to have the default enable FID.

@lucidrains You mentioned above that you would like to have save and load methods on the trainer. It would be convenient for me to have those methods available now so that the script is simpler and future compatible. If you are not planning to implement those in the near future, I would like to add my save and load methods with signatures like `trainer.save(file_path)` and `trainer.load(file_path)` as mostly functional placeholders. That sound like a plan to you?

rom1504:
I'd recommend splitting this in many PRs to help getting things merged sooner:
* A data loader PR
* An experiment tracker PR
* A load/save PR
* A minimal training script PR

They can all be submitted independently and merged independently
At the end you'll have back ported all changed from here to master and it will be possible to close this PR

lucidrains:
@Veldrovive thanks for breaking up the PRs! i took a quick look and merged in two of the more recent blockers. my hands are still a bit tied next week, and in the event that I take up any new opportunities, just wanted to let you know I gave Romain collaborator status (as he is the most senior level) He can merge in the rest of the PR as he sees fit :)

lucidrains:
@Veldrovive thanks for breaking up this PR! i took a quick look and merged in two of the more recent smaller blocker PRs. my hands are still a bit tied next week, and in the event that I take up any new opportunities, just wanted to let you know I gave Romain collaborator status (as he is the most senior level) He can merge in the rest of the PR as he sees fit :)

lucidrains:
@Veldrovive thanks for breaking up this PR! i took a quick look and merged in two of the more recent smaller blocker PRs. my hands are still a bit tied next week, and in the event that I take up any new opportunities, just wanted to let you know I gave Romain collaborator status (as he is the most senior) He can merge in the rest of the PR as he sees fit :)

Veldrovive:
All changes merged in separate pull requests.

rom1504:
Nice work @Veldrovive !

PR #92, opened 2022-05-13
Added projected gan as head alternative

zimonitrome:
Lightweight gan is a great proposal. But projected gan seemingly beats lightweight gan in speed (here referred to as FastGAN)
image

Maybe combining both could be useful.

Added links to both proposals in the readme.


lucidrains:
@zimonitrome oh hey, yes, i'm aware of projected GAN, but i think its probably better to stick with what nvidia has, save for a few minor modifications

thanks for the suggestion though!

PR #94, opened 2022-05-13
Add Text Encoding For OpenAIClipAdapter

nousr:
This just updates the `OpenAIClipAdapter.embed_text()` function to essentially override `CLIP.encode_text()` . This allows us to grab the 🙈-state before computing the actual embedding.

I don't know of any fancy tricks to extract that specific part of the network without doing this, so afaik this is the cleanest way!

### DIFF
--- a/dalle2_pytorch/dalle2_pytorch.py
+++ b/dalle2_pytorch/dalle2_pytorch.py
@@ -262,9 +262,21 @@ class OpenAIClipAdapter(BaseClipAdapter):
         text_mask = text != 0
         assert not self.cleared
 
-        text_embed = self.clip.encode_text(text)
-        text_encodings = self.text_encodings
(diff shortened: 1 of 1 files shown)


lucidrains:
@nousr ohh hey, sorry if you spent all day coding this up, but i already have a solution for this using a `register_forward_hook`

https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/dalle2_pytorch.py#L248

lucidrains:
@nousr check with me next time before you spend a whole day doing engineering lol

nousr:
Oh not at all, it took like two minutes haha. 

I saw the hook but wasn't sure if that's what it was for. 

How would you use the hook in this context? @lucidrains

PR #95, opened 2022-05-14
Migrate to text-conditioned prior training

nousr:
removed text embedding reader and opted to compute encodings and embeddings at train-time.

& some defaults changes pertaining to the switch, as well as making the default model larger (per Katherine's anecdotal evidence)

### DIFF
--- a/dalle2_pytorch/dataloaders/embedding_wrapper.py
+++ b/dalle2_pytorch/dataloaders/embedding_wrapper.py
@@ -173,7 +173,7 @@ def make_splits(
             text_reader=text_reader,
             batch_size=batch_size,
             start=eval_stop,
-            stop=num_data_points,
+            stop=int(num_data_points),
             device=device,
         )
 
--- a/dalle2_pytorch/dataloaders/embedding_wrapper.py
+++ b/dalle2_pytorch/dataloaders/embedding_wrapper.py
@@ -89,19 +89,23 @@ def make_splits(
 
     assert img_url is not None, "Must supply some image embeddings"
 
-    # compute split points
-    train_set_size = int(train_split * num_data_points)
-    eval_set_size = int(eval_split * num_data_points)
-    eval_stop = train_set_size + eval_set_size
-
     if text_conditioned:
         assert meta_url is not None, "Must supply metadata url if text-conditioning"
         image_reader = EmbeddingReader(
-            img_url,
+            embeddings_folder=img_url,
             file_format="parquet_npy",
             meta_columns=["caption"],
             metadata_folder=meta_url,
         )
+
+        # compute split points
+        if num_data_points > image_reader.count:
+            print("Specified point count is larger than the number of points available...defaulting to max length of reader.")
+            num_data_points = image_reader.count
+
+        train_set_size = int(train_split * num_data_points)
+        eval_set_size = int(eval_split * num_data_points)
+        eval_stop = int(train_set_size + eval_set_size)
 
         train_loader = PriorEmbeddingLoader(
             text_conditioned=text_conditioned,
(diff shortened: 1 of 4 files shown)


train_diffusion_prior.py
```diff
@@ -154,10 +157,10 @@ def train(image_embed_dim,
...
 
     # Get image and text embeddings from the servers
-    print_ribbon("Downloading embeddings - image and text")
-    image_reader = EmbeddingReader(embeddings_folder=image_embed_url, file_format="npy")
-    text_reader  = EmbeddingReader(embeddings_folder=text_embed_url, file_format="npy")
-    num_data_points = text_reader.count
+    print_ribbon("Downloading Image Embeddings")
+    image_reader = EmbeddingReader(embeddings_folder=image_embed_url,
+                                   metadata_folder=meta_url, meta_columns=['caption'], file_format="parquet_npy")
+    num_data_points = image_reader.count
```
nousr, line 163: @rom1504 This is the biggest change, moving to a single EmbeddingReader. 

Given how much better Katherine's results are with text-conditioning I feel as though it is a safe bet to drop support for the embedding-only training scheme we had previously.

rom1504:
I would advise strongly to keep the option to train with only precomputed text and image embeddings

It's good to add the option to use the text encodings but I think it's important to keep the precomputed only option

nousr:
@rom1504 @lucidrains 

I made a wrapper that abstracts away the logic to handle both `embedding-only` and `text_conditioned+image_embedding`. 

Now it's a lot cleaner as you just call `for img, text in dataloader:` and it will automatically handle returning either the embeddings or tokenized text.

`make_splits()` will return a embedding wrapper for each of the requested splits (train/eval/test) which enables us to just pass around a single variable instead of `num_data_points, starts, stops` (etc).

lucidrains:
@nousr ohhh, code looks 💯 💯  :)

rom1504:
much nicer indeed!

does it work? can you share a wandb run?

nousr:
> much nicer indeed!
> 
> does it work? can you share a wandb run?

@rom1504 
Let's find out! 🚀  Just launched these.
https://wandb.ai/nousr_laion/new-train-script?workspace=user-nousr_laion

train_diffusion_prior.py
```diff
@@ -363,6 +363,7 @@ def train(
                 ### Evaluate model(validation run) ###
                 eval_model(diffusion_prior, eval_loader, dp_condition_on_text_encodings, dp_loss_type, phase="Validation")
 
+            step += 1
```
nousr, line 366: small bug here

dalle2_pytorch/dataloaders/embedding_wrapper.py
```diff
@@ -20,7 +20,7 @@ def __init__(
         self.text_conditioned = text_conditioned
 
         if not self.text_conditioned:
-            self.text_reader = image_reader
+            self.text_reader = text_reader
```
nousr, line 23: small bug here: fixed

nousr: maybe a little bigger than small 😅 hahah

nousr:
So @rom1504  the only thing I notice is that `NUM_TEST_EMBEDDINGS` is no longer used.

That means validation loops currently take much longer, should the default val-split just be very small? Or should I add a check inside the validation loop to break when we check more than specified.

rom1504:
yes the validation scores computed often should be on a small part of the val set

nousr:
> yes the validation scores computed often should be on a small part of the val set

alright @rom1504  I went ahead and just pushed some smaller values. this will at least let us prototype while we experiment with the conditioned prior. 

They're more or less arbitrary at this point (and ofc modifiable in the CLI)...I figure when we have some idea of how the prior trains we can go and fine-tune the values for val sets and the tracking intervals

lucidrains:
lgtm!

PR #105, opened 2022-05-20
Update decoder dataloader

Veldrovive:
Fixed a few bugs with the file shard width, improved error handling to catch errors early on and only return good data, and added a readme file for the data loaders so we do not clutter the main readme with irrelevant information.


Veldrovive:
This is currently blocking the decoder trainer as it relies on these updates to work.

PR #106, opened 2022-05-20
Implemented the wandb tracker

Veldrovive:
Introduced three new methods to the trackers to make experiments agnostic to the tracker.

`log_images`: Interprets a tensor as an image for the purpose of logging. This could also be achieved with something like an `image=True` parameter on the normal log method, but since the functionality is basically entirely different I thought it was more appropriate to have a separate method. A more elegant solution for wandb may be to assume all images will be PIL images in which case transformations could be applied only to those, but that would add lots of complication with trackers that handle images in an entirely different way from numerical logs.

`save_state_dict`: This takes a state dict instead of an object that has a `state_dict` method because in some cases you want to manipulate the state dict before saving it and the only added difficulty is that you have to call this method as `trainer.save_state_dict(model.state_dict())`.

`recall_state_dict`: In some cases, you may be tracking using a service like wandb but want to load a local checkpoint. This means it makes the most sense for the load method to take a source that decides which load function to use and pass through all other parameters.

Finally, I added a `base_path` parameter which tells the tracker where to store any local data. Since most trackers also store some data on disc I added this as a parameter to all trackers.

### DIFF
--- a/dalle2_pytorch/trackers.py
+++ b/dalle2_pytorch/trackers.py
@@ -1,4 +1,6 @@
 import os
+from itertools import zip_longest
+from enum import Enum
 import torch
 from torch import nn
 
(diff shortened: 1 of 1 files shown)


Veldrovive:
This is currently blocking the decoder trainer as it relies on these methods to work.

The diffusion prior trainer would also need a small change to provide a `base_path` to its tracker.

lucidrains:
looks great, thanks for taking on the tracker agnostic code :)

PR #112, opened 2022-05-23
Distributed Training of the Prior

nousr:
I went ahead and started a branch to facilitate distributed training of the prior.

**Hugging Face Accelerator** should allow everyone to run and contribute to the same script--only needing to run external `accelerate config` and `accelerate launch train.py` commands when training in a distributed manner. This saves us from trying to figure out the best way to setup boilerplate that supports all users. It also has built-in trackers for WANDB (and others), and an extensible interface for writing your own, which is handy. 

---

I've confirmed basic functionality using these two initial commits, and have a rough checklist for things that probably need to get done before merging:

Completed: ✅ 
- [x] Make the embedding wrapper fit the actual PyTorch API for use in distributed applications [moving to seperate PR]
- [x] Verify users can train multi-gpu and single-gpu from same script
- [x] Enable the use of config.json files for the prior [probably seperate PR]
- [x] Ensure speed is comparable to *out-of-the-box* D.D.P
- [x] Implement checkpoint & restarts (_note: seems to still need `group_wd_params=False` (even after deterministic fix) in limited testing_)
- [x] Verify stability during long training sessions
- [x] Ensure metrics are tracked properly
- [x] Update train_diffusion_prior.py
- [x] ensure the user doesn't **need** to use accelerator if they don't want to
- [x] Bring back in EMA support

### DIFF
--- a/dalle2_pytorch/trainer.py
+++ b/dalle2_pytorch/trainer.py
@@ -14,6 +14,8 @@ from dalle2_pytorch.optimizer import get_optimizer
 from dalle2_pytorch.version import __version__
 from packaging import version
 
+from accelerate import Accelerator
+
 import numpy as np
 
 # helper functions
@@ -189,13 +191,13 @@ class EMA(nn.Module):
     By adjusting the power, you can control how fast EMA will ramp up to your specified beta.
 
     @crowsonkb's notes on EMA Warmup:
-    
+
     If gamma=1 and power=1, implements a simple average. gamma=1, power=2/3 are
     good values for models you plan to train for a million or more steps (reaches decay
     factor 0.999 at 31.6K steps, 0.9999 at 1M steps), gamma=1, power=3/4 for models
     you plan to train for less (reaches decay factor 0.999 at 10K steps, 0.9999 at
     215.4k steps).
-    
+
     Args:
         inv_gamma (float): Inverse multiplicative factor of EMA warmup. Default: 1.
         power (float): Exponential factor of EMA warmup. Default: 1.
@@ -205,7 +207,7 @@ class EMA(nn.Module):
         self,
         model,
         beta = 0.9999,
-        update_after_step = 10000,
+        update_after_step = 100,
         update_every = 10,
         inv_gamma = 1.0,
         power = 2/3,
@@ -280,6 +282,7 @@ class EMA(nn.Module):
     def __call__(self, *args, **kwargs):
         return self.ema_model(*args, **kwargs)
 
+
 # diffusion prior trainer
 
 def prior_sample_in_chunks(fn):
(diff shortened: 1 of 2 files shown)


nousr:
@lucidrains @rom1504 taking this out of WIP. 

There's two small `#TODO`'s, but they're not specific to this feature _(aka: can be a seperate PR to update the config module)_.

I figure we might as well discuss the changes to the trainer API and training code layout. That way we can get _**main**_ caught up with the advancements i've kept to myself over the past two weeks haha

dalle2_pytorch/trainer.py
```diff
@@ -14,6 +14,8 @@
 from dalle2_pytorch.version import __version__
 from packaging import version
 
+from accelerate import Accelerator
```
nousr, line 17: do we want to lazy import this? or should we depend on HFA?

dalle2_pytorch/trainer.py
```diff
@@ -303,88 +306,189 @@ def __init__(
...
+            return torch.nn.utils.clip_grad_norm_(*args)
+
+    def backprop(self, x):
+        if exists(self.accelerator):
+            self.accelerator.backward(x)
+        else:
+            try:
+                x.backward()
+            except Exception as e:
+                self.print(f"Caught error in backprop call: {e}")
```
nousr, line 398: how do we feel about these wrappers? @Veldrovive adopted the style and it seems to fit nicely, but I'm totally open to alternatives

dalle2_pytorch/trainer.py
```diff
@@ -399,17 +503,32 @@ def update(self):
...
+            return self.diffusion_prior.sample(*args, **kwargs)
 
     @torch.no_grad()
     def sample_batch_size(self, *args, **kwargs):
-        return self.ema_diffusion_prior.ema_model.sample_batch_size(*args, **kwargs)
+        if self.use_ema:
+            return self.ema_diffusion_prior.ema_model.sample_batch_size(*args, **kwargs)
+        else:
+            return self.diffusion_prior.sample_batch_size(*args, **kwargs)
+
```
nousr, line 526: added support for non-ema sampling

dalle2_pytorch/trainer.py
```diff
@@ -399,17 +503,32 @@ def update(self):
...
+    @torch.no_grad()
+    @cast_torch_tensor
+    @prior_sample_in_chunks
+    def embed_text(self, *args, **kwargs):
+        return self.unwrap_model(self.diffusion_prior).clip.embed_text(*args, **kwargs)
```
nousr, line 531: helper function to easily access the internal clip

nousr (commented):
here are the major changes to `trainer.py`

lucidrains:
👏👏 tremendous work!

dalle2_pytorch/trainer.py
lucidrains: i think we should depend on it 😄 i think it should have a good future!

dalle2_pytorch/trainer.py
lucidrains: i think it would be slightly cleaner to have it the way the tracker is handled https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/trackers.py#L32 , but definitely not a showstopper!

dalle2_pytorch/trainer.py
lucidrains: comes down to personal preference, but i'd like

python
```
model = self.ema_diffusion_prior.ema_model if self.use_ema else self.diffusion_prior
return model.sample(*args, **kwargs)
```

dalle2_pytorch/trainer.py
lucidrains: drop a thank you to huggingface too! in particular Sylvain for his awesome engineering work :smile:

nousr: love that, will do!

PR #113, opened 2022-05-24
Add basic config support for the prior

nousr:
Using this I was able to successfully call `config.prior.create()` and get a `DiffusionPrior` object back using something like this,,,

```python
def validate_prior(config):
    # try loading prior form config
    prior = config.prior.create()
    assert type(prior) is DiffusionPrior, "Failed loading prior!"

@click.command()
@click.option("--config_file", default="DALLE2-pytorch/configs/train_prior_config.example.json", help="Path to config file")
def main(config_file):
    print("Recalling config from {}".format(config_file))
    config = TrainDiffusionPriorConfig.from_json_path(config_file)

    # run quick test
    validate_prior(config)

if __name__ == "__main__":
    main()
```


lucidrains:
yesss, this is excellent! 💯 🚀 !

dalle2_pytorch/train_configs.py
```diff
@@ -16,7 +16,43 @@ def default(val, d):
...
+
+class AdapterConfig(BaseModel):
+    make: str = "openai"
+    model: str = "ViT-L/14"
+
+    def create(self):
+        if self.make == "openai":
+            return OpenAIClipAdapter(self.model)
+        elif self.make == "x-clip":
+            return XClipAdapter(self.model)
```
lucidrains, line 51: the factory may need to take in some `base_model_kwargs` and splat it into `XClip`

dalle2_pytorch/train_configs.py
```diff
@@ -46,14 +86,51 @@ class DiffusionPriorConfig(BaseModel):
     loss_type: str = 'l2'
     predict_x_start: bool = True
     beta_schedule: str = 'cosine'
+    condition_on_text_encodings: bool = True
+
+    class Config:
+            extra = "allow"
```
lucidrains, line 92: nit: one indentation too many

dalle2_pytorch/train_configs.py
```diff
@@ -16,7 +16,43 @@ def default(val, d):
...
+    test: float = 0.1
+
+    @root_validator
+    def validate_all(cls, fields):
+        actual_sum = sum([*fields.values()])
+        if actual_sum != 1.:
+            raise ValueError(f'{fields.keys()} must sum to 1.0. Found: {actual_sum}')
+        return fields
+
+class TrackerConfig(BaseModel):
```
lucidrains, line 33: tracker config could assume the responsibility of instantiating either the console or wandb trackers at some point, but could also be another PR

dalle2_pytorch/train_configs.py
nousr: Should `base_model_kwargs` be very explicit? and only for XClip?
```python
class AdapterConfig(BaseModel):
    make: str = "openai"
    model: str = "ViT-L/14"
    base_model_kwargs: Dict[str, Any] = None

    def create(self):
        if self.make == "openai":
            return OpenAIClipAdapter(self.model)
        elif self.make == "x-clip":
            return XClipAdapter(self.model, **self.base_model_kwargs)
        elif self.make == "cocoa":
            return CoCaAdapter(self.model)
        else:
            raise AttributeError("No adapter with that name is available.")
```

Or (perhaps a little less pydantic), allow for extra configs and trust the user will pass meaningful values...
```python
class AdapterConfig(BaseModel):
    make: str = "openai"
    model: str = "ViT-L/14"

    class Config:
        extra = "allow"

    def create(self):
        kwargs = self.dict()
        make = kwargs.pop("make")
        model = kwargs.pop("model")

        if make == "openai":
            return OpenAIClipAdapter(model, **kwargs)
        elif make == "x-clip":
            return XClipAdapter(model, **kwargs)
        elif make == "cocoa":
            return CoCaAdapter(model, **kwargs)
        else:
            raise AttributeError("No adapter with that name is available.")
```

dalle2_pytorch/train_configs.py
nousr: nice catch, i swore i checked like 5 times for indentations and trailing whitespace 😅 😆

dalle2_pytorch/train_configs.py
nousr: Yeah totally agree! Actually, the only reason I didn't do this was because I've been experimenting with Hugging Face Accelerator and they have a built-in WANDB tracker that I've been testing out....not sure I like how restrictive their tracker is, though. (off topic: but hfa has been pretty nice, otherwise, at least so far!)

If you (or someone else) doesn't around to this, perhaps I can just do a seperate PR once I get the distributed training of the prior more fleshed out?

dalle2_pytorch/train_configs.py
lucidrains: yea, it is tricky because `XClipAdapter` actually wraps a `XClip` model, (and similarly `CoCaAdapter` wraps `Coca`) perhaps it is best to just leave them unsupported for now

dalle2_pytorch/train_configs.py
lucidrains: yup, we can always leave this for later!

dalle2_pytorch/train_configs.py
nousr: I think I see what you were getting at now, I believe I fixed it! 

If that still isn't what you had in mind, lmk, I can revert and we can just leave it unsupported for now

nousr:
@lucidrains caught a small typo, but hopefully it's all good now? https://github.com/lucidrains/DALLE2-pytorch/commit/e5222b710f728bffaf4194e25a42d24f8154ae7f

lucidrains:
yolo! thank you @nousr !

PR #116, opened 2022-05-25
[FIX #115] Use deterministic optimizer parameters

### LINKED ISSUE #115: Optimizer error when resuming from state-dict
Hey, this one has me a little stumped.

**problem**:  DiffusionPriorTrainer is not able to successfully reload from a checkpoint and continue training.

calling `.save()` and then `.load()` seems to corrupt the optimizer in some way. here is a minimal code example that should reproduce the error on main.

(steps to reproduce are simple and in the file as a docstring)
https://gist.github.com/nousr/5f6877c2bcb8b0c1b806ed8e206d39af

my hunch is that it has something to do with the `param_groups` as setting weight_decay to 0 in the trainer (i.e. using ADAM instead of ADAMW) will fix the issue. this seems to be a related issue https://github.com/pytorch/pytorch/issues/40769 with potentially the most relevant reply being:

>...It seems like some important aspect of your model changes between checkpoints. From the error, it looks like the shape of the gradient average does not match the shape of the current gradient....

Other than just skipping the re-load of the optimizer all together I haven't really been able to solve this issue fully. Does anyone have any ideas?

nousr:
caught this in the PyTorch source code when digging a little deeper into #115 

> **warning:** Parameters need to be specified as collections that have a deterministic
ordering that is consistent between runs. **Examples of objects that don't
satisfy those properties are ***`sets`*** and iterators over values of dictionaries.**

---

If there's concern that there are a lot of duplicate key/value pairs in the model parameters, perhaps we can look into some other deterministic iterable that's both deterministic and unique, but this seems to solve the issue that I was running into...

### DIFF
--- a/dalle2_pytorch/optimizer.py
+++ b/dalle2_pytorch/optimizer.py
@@ -21,8 +21,6 @@ def get_optimizer(
     if wd == 0:
         return Adam(params, lr = lr, betas = betas, eps = eps)
 
-    params = set(params)
-
     if group_wd_params:
         wd_params, no_wd_params = separate_weight_decayable_params(params)
 


lucidrains:
@nousr boss

PR #117, opened 2022-05-26
bibtex typo for max_vit

vztu:
Hi @lucidrains, thanks for referencing max_vit!! But found this typo again =). 

Just wondering if the `grid attention` you posted in the to-do list is from our paper? If so, I'd sincerely appreciate it if you can add the arxiv paper link in the item just like other items~ 
https://arxiv.org/abs/2204.01697

### DIFF
--- a/README.md
+++ b/README.md
@@ -1140,7 +1140,7 @@ This library would not have gotten to this working state without the help of
 ```bibtex
 @inproceedings{Tu2022MaxViTMV,
     title   = {MaxViT: Multi-Axis Vision Transformer},
-    author  = {Zhe-Wei Tu and Hossein Talebi and Han Zhang and Feng Yang and Peyman Milanfar and Alan Conrad Bovik and Yinxiao Li},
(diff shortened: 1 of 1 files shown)


lucidrains:
@vztu oh hey! yup, should be fixed, must be a bug with semantic scholar :pray: 

do you happen to know if grid attention works with generative models well?

lucidrains:
@vztu ohh yes, that's right! now i remembered why i reached for your paper (then forgot why lol) - in these recent DDPM works, they often remove attention at the higher stages, but i'm going to try out grid attention

thanks again!

vztu:
Yes removing attention in higher stages makes sense as it requires a lot of computation; check this [HiT paper](https://arxiv.org/abs/2106.07631) as an example. Also, I would assume adding global attention in high-res stages would have marginal gains as well... But at least our grid attention (+block attention) allows you to have such a try and it's up to your compute budge to decide up to which stage you want to retain global interaction to balance performance vs complexity.

Again, thank you very much for considering our work!! :)

PR #121, opened 2022-05-28
Distributed Training of the Decoder

Veldrovive:
The same methods used in #112 are used here as well. Implementation is via [huggingface accelerate](https://github.com/huggingface/accelerate) to minimize changes needed to have a script capable of single or multi GPU training.

The method used was to convert the decoder trainer class to wrap its models with an accelerator object passed in as an argument. Functionality remains mostly unchanged. One new change was made to separate the loss values for the different unets during training and validation in order to make it easier to see which need more training.

This is not ready due to some remaining issues, but I wanted a place to track them and have others provide input.
Known Issues:
- [x] Evalaution metrics hang on distributed mode
- [x] Training loss drops slowly and arbitrarily diverges (This appears to also be an issue on the main branch)
- [x] Loading the model causes the optimizer to error

Roadmap:
- [x] Implement distributed `DecoderTrainer`
- [x] Update `train_decoder.py` to use the distributed trainer.
- [x] Fix remaining issues
- [x] Verify functionality on large training set
- [ ] Port script for imagen

### DIFF
--- a/dalle2_pytorch/dalle2_pytorch.py
+++ b/dalle2_pytorch/dalle2_pytorch.py
@@ -2099,7 +2099,8 @@ class Decoder(BaseGaussianDiffusion):
         text_encodings = None,
         batch_size = 1,
         cond_scale = 1.,
-        stop_at_unet_number = None
+        stop_at_unet_number = None,
+        distributed = False,
     ):
         assert self.unconditional or exists(image_embed), 'image embed must be present on sampling from decoder unless if trained unconditionally'
 
@@ -2118,7 +2119,7 @@ class Decoder(BaseGaussianDiffusion):
 
         for unet_number, unet, vae, channel, image_size, predict_x_start, learned_variance in tqdm(zip(range(1, len(self.unets) + 1), self.unets, self.vaes, self.sample_channels, self.image_sizes, self.predict_x_start, self.learned_variance)):
 
-            context = self.one_unet_in_gpu(unet = unet) if is_cuda else null_context()
+            context = self.one_unet_in_gpu(unet = unet) if is_cuda and not distributed else null_context()
 
             with context:
                 lowres_cond_img = None
--- a/train_decoder.py
+++ b/train_decoder.py
@@ -222,7 +222,7 @@ def recall_trainer(tracker, trainer, recall_source=None, **load_config):
     """
     Loads the model with an appropriate method depending on the tracker
     """
-    print(print_ribbon(f"Loading model from {recall_source}"))
+    trainer.accelerator.print(print_ribbon(f"Loading model from {recall_source}"))
     local_filepath = tracker.recall_file(recall_source, **load_config)
     state_dict = trainer.load(local_filepath)
     return state_dict["epoch"], state_dict["validation_losses"]
@@ -254,7 +254,6 @@ def train(
     trainer = DecoderTrainer(
         accelerator,
         decoder,
-        group_wd_params=True,
         **kwargs
     )
 
@@ -265,6 +264,7 @@ def train(
 
     if exists(load_config) and exists(load_config.source):
         start_epoch, validation_losses = recall_trainer(tracker, trainer, recall_source=load_config.source, **load_config.dict())
+        accelerator.print(f"Loaded model from {load_config.source} on epoch {start_epoch} with minimum validation loss {min(validation_losses)}")
     trainer.to(device=inference_device)
 
     if not exists(unet_training_mask):
@@ -431,7 +431,7 @@ def create_tracker(accelerator, config, tracker_type=None, data_path=None):
     data_path = data_path or tracker_config.data_path
     tracker_type = tracker_type or tracker_config.tracker_type
 
-    if tracker_type is None:
+    if tracker_type is "dummy":
         tracker = DummyTracker(data_path)
         tracker.init(**init_config)
     elif tracker_type == "console":
@@ -469,7 +469,6 @@ def initialize_training(config):
     shards_per_process = len(all_shards) // world_size
     assert shards_per_process > 0, "Not enough shards to split evenly"
     my_shards = all_shards[rank * shards_per_process: (rank + 1) * shards_per_process]
-
     dataloaders = create_dataloaders (
         available_shards=my_shards,
         img_preproc = config.data.img_preproc,
@@ -483,12 +482,13 @@ def initialize_training(config):
     # Create the decoder model and print basic info
     decoder = config.decoder.create()
     num_parameters = sum(p.numel() for p in decoder.parameters())
-    accelerator.print(f"Number of parameters: {num_parameters}")
 
     # Create and initialize the tracker if we are the master
-    tracker = create_tracker(accelerator, config) if rank == 0 else create_tracker(accelerator, config, tracker_type=None)
+    tracker = create_tracker(accelerator, config) if rank == 0 else create_tracker(accelerator, config, tracker_type="dummy")
 
     accelerator.print(print_ribbon("Loaded Config", repeat=40))
+    accelerator.print(f"Running training with {accelerator.num_processes} processes and {accelerator.distributed_type} distributed training")
+    accelerator.print(f"Number of parameters: {num_parameters}")
     train(dataloaders, decoder, accelerator,
         tracker=tracker,
         inference_device=accelerator.device,
--- a/train_decoder.py
+++ b/train_decoder.py
@@ -117,7 +117,6 @@ def get_example_data(dataloader, device, n=5):
         captions.extend(list(txt))
         if len(images) >= n:
             break
-    print("Generated {} examples".format(len(images)))
     return list(zip(images[:n], embeddings[:n], captions[:n]))
 
 def generate_samples(trainer, example_data, text_prepend=""):
@@ -161,21 +160,25 @@ def evaluate_trainer(trainer, dataloader, device, n_evaluation_samples=1000, FID
     # Convert from [0, 1] to [0, 255] and from torch.float to torch.uint8
     int_real_images = real_images.mul(255).add(0.5).clamp(0, 255).type(torch.uint8)
     int_generated_images = generated_images.mul(255).add(0.5).clamp(0, 255).type(torch.uint8)
+
+    def null_sync(t, *args, **kwargs):
+        return [t]
+
     if exists(FID):
-        fid = FrechetInceptionDistance(**FID)
+        fid = FrechetInceptionDistance(**FID, dist_sync_fn=null_sync)
         fid.to(device=device)
         fid.update(int_real_images, real=True)
         fid.update(int_generated_images, real=False)
         metrics["FID"] = fid.compute().item()
     if exists(IS):
-        inception = InceptionScore(**IS)
+        inception = InceptionScore(**IS, dist_sync_fn=null_sync)
         inception.to(device=device)
         inception.update(int_real_images)
         is_mean, is_std = inception.compute()
         metrics["IS_mean"] = is_mean.item()
         metrics["IS_std"] = is_std.item()
     if exists(KID):
-        kernel_inception = KernelInceptionDistance(**KID)
+        kernel_inception = KernelInceptionDistance(**KID, dist_sync_fn=null_sync)
         kernel_inception.to(device=device)
         kernel_inception.update(int_real_images, real=True)
         kernel_inception.update(int_generated_images, real=False)
@@ -186,10 +189,21 @@ def evaluate_trainer(trainer, dataloader, device, n_evaluation_samples=1000, FID
         # Convert from [0, 1] to [-1, 1]
         renorm_real_images = real_images.mul(2).sub(1)
         renorm_generated_images = generated_images.mul(2).sub(1)
-        lpips = LearnedPerceptualImagePatchSimilarity(**LPIPS)
+        lpips = LearnedPerceptualImagePatchSimilarity(**LPIPS, dist_sync_fn=null_sync)
         lpips.to(device=device)
         lpips.update(renorm_real_images, renorm_generated_images)
         metrics["LPIPS"] = lpips.compute().item()
+
+    if trainer.accelerator.num_processes > 1:
+        # Then we should sync the metrics
+        metrics_order = sorted(metrics.keys())
+        metrics_tensor = torch.zeros(1, len(metrics), device=device, dtype=torch.float)
+        for i, metric_name in enumerate(metrics_order):
+            metrics_tensor[0, i] = metrics[metric_name]
+        metrics_tensor = trainer.accelerator.gather(metrics_tensor)
+        metrics_tensor = metrics_tensor.mean(dim=0)
+        for i, metric_name in enumerate(metrics_order):
+            metrics[metric_name] = metrics_tensor[i].item()
     return metrics
 
 def save_trainer(tracker, trainer, epoch, step, validation_losses, relative_paths):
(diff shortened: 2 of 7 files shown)


dalle2_pytorch/dalle2_pytorch.py
```diff
@@ -2057,7 +2058,7 @@ def sample(
 
         for unet_number, unet, vae, channel, image_size, predict_x_start, learned_variance in tqdm(zip(range(1, len(self.unets) + 1), self.unets, self.vaes, self.sample_channels, self.image_sizes, self.predict_x_start, self.learned_variance)):
 
-            context = self.one_unet_in_gpu(unet = unet) if is_cuda else null_context()
+            context = self.one_unet_in_gpu(unet = unet) if is_cuda and not distributed else null_context()
```
Veldrovive, line 2061: One downside of using accelerate instead of handling ddp at a low level is that accelerate does not like when submodels are modified in any way. This becomes an issue in the sample method when a unet is moved to a device. In order to get around this, we create a parameter that prevents this move if we are doing distributed training.

nousr:
> Evalaution metrics hang on distributed mode

If you haven't already figured this one out, you might need to unwrap the model before doing evaluation. Not doing so can cause a deadlock.

> Loading the model causes the optimizer to error

Could be an issue with param groups, I had a lot of issues with loading the optimizer. 1ed0f9d80b097594e6294bad432d06ce3fdcb220 and if you can, try using `group_wd_params=False`, it seems to work for the prior, at least. (Note: we can try using an `OrderedDict` for grouping the `wd_params`, I haven't looked into it, but it may be more deterministic than the standard dictionary.

Veldrovive:
The hanging on evaluation problem was caused by torchmetrics assuming evaluation is happening on all processes while I was only doing it on master. Using torchmetrics built in distributed mode causes errors with further dataloading, however. So one rabbit hole later I found that it was a problem with how torchmetrics does its gather and substituting my own gather using accelerate fixes the issue.

Veldrovive:
> Could be an issue with param groups, I had a lot of issues with loading the optimizer. [1ed0f9d](https://github.com/lucidrains/DALLE2-pytorch/commit/1ed0f9d80b097594e6294bad432d06ce3fdcb220) and if you can, try using `group_wd_params=False`, it seems to work for the prior, at least. (Note: we can try using an `OrderedDict` for grouping the `wd_params`, I haven't looked into it, but it may be more deterministic than the standard dictionary.

@nousr Yea, that seems to have fixed the issue. I don't have a good handle on why the split results in non-deterministic behavior though. Is it just the conversion between sets and lists that is messing up the parameter order?

nousr:
>  Is it just the conversion between sets and lists that is messing up the parameter order?

@Veldrovive oh shoot! yeah....good catch--I didn't even see those `set()` calls up there in `separate_weight_decayable_params()`. I was so hyperfocused on `get_optimizer()` that once it worked with `group_wd_params=False` I was happy enough to continue on with my life😆 

I bet changing those to `list()` would fix the issue, should be an easy test & PR (assuming it works) if you get some time.

Veldrovive:
The optimizer has been fixed and tested in #130 so it should no longer be necessary to use the `group_wd_params=False` option.

Veldrovive:
I brought the branch up with main and started a longer term test run with only the initial unet on 3 GPUs. You can track its progress [here](https://wandb.ai/veldrovive/dalle2_train_decoder/runs/1qywwsxb?workspace=user-veldrovive).

Veldrovive:
The training run above converged about as quickly as we would expect given its distributed batch size and all evaluation, saving, and logging works.

Veldrovive:
The change to model saving could be part of its own pull request for improving training features as it does not really fit under the umbrella of distributed training. I also want to change config logging to save the config dict as a file instead of in the wandb config object as that is more general so I could open a pull request for those features if we want to be good about keeping separate feature changes in separate requests. Up to whoever's merging if I should separate it out. I just needed the feature for testing in this branch.

I can also go and squash these commits if you do not do a squash during merge.

lucidrains:
> The training run above converged about as quickly as we would expect given its distributed batch size and all evaluation, saving, and logging works.

with accelerate, would we still have the option to go single GPU if need be? (given the big changes to the trainer class)

Veldrovive:
> with accelerate, would we still have the option to go single GPU if need be? (given the big changes to the trainer class)

That's actually my favorite feature of accelerate. If you launch with `python train_decoder.py` you get bog standard training on the first available device and accelerate basically does nothing. You do have to abide by the rules of accelerate, however, which really only means it is harder to select which device to train on since you have to use the `CUDA_VISIBLE_DEVICES` environment variable.

The simplicity does come with the cost that in order to use extra features like mixed precision you have to use `accelerate launch` since accelerate handles scaling itself, but it works so well in every situation that it feels worth the slight inconvenience.

lucidrains:
very cool and great job! i'll review it this weekend and we can try to have it merged on Monday :)

1073521013:
> The training run above converged about as quickly as we would expect given its distributed batch size and all evaluation, saving, and logging works.

First of all, thanks very much about the contribution of your distributed training code. When running this decoder distributed training,   it will hang up in the line of **self.accelerator.backward(loss)** if make **find_unused_parameters=True**, it will hang up in the line of **loss_map = {f"Unet {index} Training Loss": loss.item() for index, loss in enumerate(unet_average_loss)
 if loss != 0}**  (The output result of **accelerator.gather(unet_losses_tensor)** cannot be iterated) if make find_unused_parameters=False. Thanks again, looking forward to your reply.

Veldrovive:
> First of all, thanks very much about the contribution of your distributed training code. When running this decoder distributed training, it will hang up in the line of **self.accelerator.backward(loss)** if make **find_unused_parameters=True**, it will hang up in the line of **loss_map = {f"Unet {index} Training Loss": loss.item() for index, loss in enumerate(unet_average_loss) if loss != 0}** (The output result of **accelerator.gather(unet_losses_tensor)** cannot be iterated) if make find_unused_parameters=False. Thanks again, looking forward to your reply.

That is very strange behavior. I assume you are editing the line `ddp_kwargs = DistributedDataParallelKwargs(find_unused_parameters=True)`?

Setting `find_unused_parameters` to False should error on the backward pass because `unets.0.null_text_embed` does not participate in the loss calculation since text encodings are not supported. Did you modify the code to use text encodings?

I don't know of anything that would cause it to hang on the dict comprehension. If it is hanging, it should be on a gather call. Can you print out the intermediate value between the gather and comprehension?

Veldrovive:
@1073521013 Ok, I don't have any idea what could be causing the issue. There are three options I see. It could be something I didn't account for that you are using in your config, a code change that doesn't seem like it should affect anything but still causes issues, or a torch/accelerate version mismatch. If you could post your config in a pastebin we can start eliminating options here.

1073521013:
> @1073521013 Ok, I don't have any idea what could be causing the issue. There are three options I see. It could be something I didn't account for that you are using in your config, a code change that doesn't seem like it should affect anything but still causes issues, or a torch/accelerate version mismatch. If you could post your config in a pastebin we can start eliminating options there.

torch/accelerate version has matched(same problem). and my config:
{
    "decoder": {
        "unets": [
            {
                "dim": 512,
                "image_embed_dim": 512,
                "cond_dim": null,
                "channels": 3,
                "dim_mults": [1, 2, 3, 4],
                "attn_dim_head": 64,
                "attn_heads": 32
            }
        ],
        "image_sizes": [64],
        "channels": 3,
        "timesteps": 1000,
        "loss_type": "l2",
        "beta_schedule": "cosine",
        "learned_variance": true
    },
    "data": {
        "webdataset_base_url": "/home/notebook/data/group/ma/laion400M_data/laion400m-met-release/laion400m-webdataset/{}.tar",
        "embeddings_url": "/home/notebook/data/group/laion_400_embeddings/laion400m-embeddings/img_emb",
        "num_workers": 2,
        "batch_size": 25,
        "start_shard": 0,
        "end_shard": 41512,
        "shard_width": 5,
        "index_width": 4,
        "splits": {
            "train": 0.9,
            "val": 0.05,
            "test": 0.05
        },
        "shuffle_train": true,
        "resample_train": false,
        "preprocessing": {
            "RandomResizedCrop": {
                "size": [64, 64],
                "scale": [0.75, 1.0],
                "ratio": [1.0, 1.0]
            },
            "ToTensor": true
        }
    },
    "train": {
        "epochs": 100,
        "lr": 2e-5,
        "wd": 0.1,
        "max_grad_norm": 0.5,
        "save_every_n_samples": 300000,
        "n_sample_images": 6,
        "device": "cuda:0",
        "epoch_samples": 2000000,
        "validation_samples": 50000,
        "use_ema": false,
        "ema_beta": 0.99,
        "amp": false,
        "save_all": false,
        "save_latest": true,
        "save_best": true,
        "unet_training_mask": [true]
    },
    "evaluate": {
        "n_evaluation_samples": 10,
        "FID": {
            "feature": 64
        },
        "IS": {
            "feature": 64,
            "splits": 10
        },
        "KID": {
            "feature": 64,
            "subset_size": 10
        },
        "LPIPS": {
            "net_type": "vgg",
            "reduction": "mean"
        }
    },
    "tracker": {
        "tracker_type": "wandb",
        "data_path": "./models_all",
        "image_path":"./wandb",
        "wandb_entity": "foreveryoung",
        "wandb_project": "decoder",
        "verbose": false
    },
    "load": {
        "source": null,
        "run_path": "foreveryoung/decoder/azotfo37",
        "file_path": "latest.pth",
        "resume": false
    }
}

Veldrovive:
Ah, ok the most likely explanation is that 400M has not been reordered so the image shards do not line up with the embeddings shards. If that happens one process will load a shard correctly and another won't which leads to a hang on the first gather it gets to. The reason it doesn't error is because webdataset hangs instead of erroring if an image shard it's trying to load doesn't exist.  This is also a problem if you use an end shard that is greater than the number of shards. This is a good thing to bring up because it will be one of the most common issues and it is worth addressing with better logging.

Logging won't solve the base issue of the dataset not working for training the decoder, though. Which source are you using to get 400M? I can probably reorder the embeddings so that it is a valid dataset.

We should also add an option to do on the fly embedding generation so that if a dataset does not have reordered embeddings and you don't have the ability to generate them yourself you can trade off speed and just load images.

1073521013:
This is indeed the case. My data is obtained from [here ](https://deploy.laion.ai/8f83b608504d46bb81708ec86e912220/). Due to some other problems, I downloaded the training data locally for training. I originally wanted to code to align it, but I saw that you have implemented embedding-dataset-reordering, so I used your script to align it. After the final alignment, I experimented with it. There are still two small problems, one is after alignment The embedding npy file has a large number of 0 matrices (which may represent images that were not successfully downloaded), resulting in lots of problems like `UserWarning: RuntimeError('Webdataset had a sample, but no embedding was found. ImgShard: 10093 - Index: 0229')`; Another problem is that although the distributed operation is running, a step only executes one batch instead of gpus*batch. It is still under research, thanks.

Veldrovive:
0 vectors are inserted into the embeddings whenever an image index was skipped when computing embeddings. That warning showing up in large quantities means a lot of image links were dead when embeddings were computed.

> Another problem is that although the distributed operation is running, a step only executes one batch instead of gpus*batch.

I am not sure exactly what this means. Are you saying that `sample` only increases by batch size every step instead of by gpus*batch size?

1073521013:
> This is indeed the case. My data is obtained from [here ](https://deploy.laion.ai/8f83b608504d46bb81708ec86e912220/). Due to some other problems, I downloaded the training data locally for training. I originally wanted to code to align it, but I saw that you have implemented embedding-dataset-reordering, so I used your script to align it. After the final alignment, I experimented with it. There are still two small problems, one is after alignment The embedding npy file has a large number of 0 matrices (which may represent images that were not successfully downloaded), resulting in lots of problems like `UserWarning: RuntimeError('Webdataset had a sample, but no embedding was found. ImgShard: 10093 - Index: 0229')`; Another problem is that although the distributed operation is running, a step only executes one batch instead of gpus*batch. It is still under research, thanks.

I carefully studied the data downloaded [before](https://deploy.laion.ai/8f83b608504d46bb81708ec86e912220/embeddings/), and found that the meta parquet data of `embeddings/metadata/` is different from the parquet log data in webdataset that the download is successful, and they are not aligned, which leads to a large number of warnings;
The second problem is that my code is wrong. and it's solved; Many thanks for your hospitality.

YUHANG-Ma:
Hi, thanks for your contribution. When I use your code, it casues an error shown in the following pic.
50af5ab87cc689a27d4a8ceceb2d88a
do you know how to fix it?

nousr:
> Hi, thanks for your contribution. When I use your code, it casues an error shown in the following pic.
> 
> 50af5ab87cc689a27d4a8ceceb2d88a
> 
> do you know how to fix it?

Hey! This issue happens when you try to call the member function of a Distributed model.

The common solution for this is "unwrapping" the model by calling `distributed_model.module.member_function()`. 

Hopefully this puts you on the right track!

YUHANG-Ma:
> 

Hi, I am also confused. Where do I need to unwrapping the model?

YUHANG-Ma:
> > Hi, thanks for your contribution. When I use your code, it casues an error shown in the following pic.
> > 50af5ab87cc689a27d4a8ceceb2d88a
> > do you know how to fix it?
> 
> Hey! This issue happens when you try to call the member function of a Distributed model.
> 
> The common solution for this is "unwrapping" the model by calling `distributed_model.module.member_function()`.
> 
> Hopefully this puts you on the right track!

image
Do you mean I need to fix it like this?

YUHANG-Ma:
> > > Hi, thanks for your contribution. When I use your code, it casues an error shown in the following pic.
> > > 50af5ab87cc689a27d4a8ceceb2d88a
> > > do you know how to fix it?
> > 
> > 
> > Hey! This issue happens when you try to call the member function of a Distributed model.
> > The common solution for this is "unwrapping" the model by calling `distributed_model.module.member_function()`.
> > Hopefully this puts you on the right track!
> 
> image Do you mean I need to fix it like this?

Thanks, I fix it!

YUHANG-Ma:
Hi, I also have an issue. My dataset contains about 17million data. But during my training process, in epoch 0 it samples more than 17million. Why did it happen?
image

YUHANG-Ma:
> > Hi, thanks for your contribution. When I use your code, it casues an error shown in the following pic. 50af5ab87cc689a27d4a8ceceb2d88a do you know how to fix it?
> 
> That's on me. It is fixed in a commit I have not not pushed.

I also met a problem that yy dataset contains about 17million data, but during my training process, in epoch 0 it samples more than 17million. It couldn't move to epoch 1 and it failed t o save the checkpoint although I set the "save_all" as true in config.

nousr:
> Hi, sorry to bother you again. I met an issue after I loaded the checkpoints. I can load the model successfully. But when it starts training, it shows an error. I am confused about it. Why did this happen even if the model is loaded successfully?
> 
> image
> 
> 

Hey! This looks like a problem with the optimizer. I would ensure that you are utilizing the same hyper-parameters for the decoder when you saved the checkpoint. Additionally, I would make sure you have the latest version of the repo.

If you've tried both of those and still get the error, please open an issue with a minimal example to recreate the problem so we can try to get to the bottom of it!

YUHANG-Ma:
> > Hi, sorry to bother you again. I met an issue after I loaded the checkpoints. I can load the model successfully. But when it starts training, it shows an error. I am confused about it. Why did this happen even if the model is loaded successfully?
> > image
> 
> Hey! This looks like a problem with the optimizer. I would ensure that you are utilizing the same hyper-parameters for the decoder as when you saved the checkpoint. Additionally, I would make sure you have the latest version of the repo (and it might be worth checking that this issue still presents itself if you save a checkpoint with the latest version as well).
> 
> If you've tried both of those and still get the error, please open an issue with a minimal example to recreate the problem so we can try to get to the bottom of it!

Thanks, I will try to download the latest version of the repo. Is that from https://github.com/Veldrovive/DALLE2-pytorch/tree/decoder_trainer_accelerate?

lucidrains:
@Veldrovive hey Aidan! thanks for continuing to push on this pull request!

we can merge this any time you think it is ready

(don't worry about Imagen btw, i will probably complete it next week solo)

Veldrovive:
@lucidrains I brought it even with main to resolve merge conflicts yesterday so that it could be merged. Nousr and I just finished tracking down a bug that I thought might have been introduced, but it turned out it was just wandb having strange behavior. [This](https://wandb.ai/veldrovive/dalle2_train_decoder/runs/3d5rytsa) is a run using the up to date version of the branch with sparse attention enabled for the unet. It is ready to be merged. Would you like me to squash the commits or jut leave it as is?

lucidrains:
@Veldrovive it is perfect! let's merge it :pray:

@nousr @Veldrovive what is both of your emails btw? could you send me an email at lucidrains@gmail.com so i can keep track?

train_decoder.py
```diff
@@ -237,157 +255,216 @@ def train(
...
+            accelerator.print(print_ribbon(f"Starting Validation {epoch}", repeat=40))
+            last_val_sample = val_sample
+            val_sample_length_tensor = torch.zeros(1, dtype=torch.int, device=inference_device)
+            average_val_loss_tensor = torch.zeros(1, trainer.num_unets, dtype=torch.float, device=inference_device)
             timer = Timer()
-            for i, (img, emb, *_) in enumerate(dataloaders["val"]):
-                sample += img.shape[0]
+            accelerator.wait_for_everyone()
+            i = 0
+            for i, (img, emb, txt) in enumerate(dataloaders["val"]):
```
mrsndmn, line 379: Shouldn't we use here prepared dataloaders from `accelerate.prepare`?

PR #122, opened 2022-05-29
Overhaul Prior DataLoader

nousr:
I've made an overhaul to the embedding data-loader for the prior. This is the updated code I've been using in my distributed training runs and I believe it is now stable enough to merge with main.

Here are the notable changes:
1. Added usage in README
2. Rename to disambiguate with the decoder's `ImageEmbeddingLoader`
4. Distributed training support
   - `make_splits` can now allocate subsets of each split to a rank of a given world size.
5. The `PriorEmbeddingDataset` *more*-properly extends the `IterableDataset`
   - it can now be wrapped with `DataLoader`
   - does not assign devices (to support distributed training)
6. Break-out creation of the `EmbeddingReader` into a seperate function (`get_reader`) to simplify API for `make_splits`.

---

_NOTE: Even though the distributed training PR is around the corner--I should still add a patch for `train_diffusion_prior.py` to be compatible with these changes...I'm getting pretty tired, though, so I'll squash that commit in the morning 😴_


lucidrains:
lgtm!

PR #130, opened 2022-05-31
Fix non deterministic optimizer creation

Veldrovive:
When using the `group_wd_params` option, the optimizer shuffles the model parameters because of conversion between sets and lists. To fix this we switch from the set difference operation to a more standard list construction.

Saving and loading the optimizer has been tested and functions after this fix.


lucidrains (approved):
@Veldrovive @nousr thank you both! :100:

Issue #131, opened 2022-05-31
Potential EMA Model Issues


nousr:
So, I've noticed a potential bug related to the exponential moving averaged priors (I haven't tested the other models).

Essentially, the EMA model seemingly refuses to "learn". I have to believe that the weights aren't being properly updated with the `online_model`, for whatever reason.

![image](https://user-images.githubusercontent.com/51308183/171259293-8967018c-dc7b-4fae-a77f-60693c5ef56f.png)

As you can see, as soon as the EMA kicks in at step 1000, the loss plateaus. I've never worked with an EMA model before, so I guess this **_could_** be expected behavior?

This issue also presents itself when running evaluation on the EMA model (the `(text<->predicted_image)` similarity should be at least 0.1 even after a few hundred steps)

![image](https://user-images.githubusercontent.com/51308183/171260256-705b5f4c-fbe6-4786-a0a8-dfdcbc87f9b8.png)

lucidrains:
it is normal i believe, you can always try lowering the `ema_beta` on the Trainer to `0.995` or even `0.99` to have it update faster

nousr:
gotcha, glad I asked then! just have to be patient then 🫡

nousr:
@lucidrains hey! when you get a chance, do you think could you explain the intuition for this line?

https://github.com/lucidrains/DALLE2-pytorch/blob/1cc288af39f171b3e7f77fdb4252682af05e17e9/dalle2_pytorch/trainer.py#L191

specifically, why the `// self.update_every`? I guess I don't see the purpose of moving `self.update_after` "up" by a factor of `update_every`?

nousr:
Re-opening this issue as @Veldrovive has confirmed that the EMA is not giving expected results for him as well. A discussion with Katherine has confirmed my suspicions as well.

Will try to do some more debugging after my class final today.

lucidrains:
@nousr oh crap, yes, you are right! i thought that i wasn't incrementing the EMA steps in accordance with the global training step, but i was

https://github.com/lucidrains/DALLE2-pytorch/commit/9cc475f6e7990b3d978128902ee0ea90614451f6 should fix the update every issue

PR #139, opened 2022-06-04
ema module fixes

nousr:
I've ironed out (#131)...it **_was_** `copy_params_from_model_to_ema`. The easy solution is to just use `.load_state_dict` but it's slower, in my testing.
image
image

---

I also took an opportunity to do some other optimizations during the decay calculation, I think this may be slightly more efficient? It seems to be the case, at least anecdotally.

**_Anecdotal Evidence_**
image

---

_NOTE: I've also went ahead and placed the beta back to 0.9999. which is what they used in the paper._

---
You can explore these charts and some more experiments during the optimization process here...
https://wandb.ai/nousr_laion/ema-test?workspace=user-nousr_laion

### DIFF
--- a/dalle2_pytorch/trainer.py
+++ b/dalle2_pytorch/trainer.py
@@ -178,7 +178,7 @@ class EMA(nn.Module):
     def __init__(
         self,
         model,
-        beta = 0.99,
+        beta = 0.9999,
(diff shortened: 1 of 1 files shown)


lucidrains:
love it! 🙏

PR #140, opened 2022-06-04
implement ema warmup from @crowsonkb

nousr:
image


lucidrains:
@nousr oh my, thank you! and thank you as always Katherine 😃  @crowsonkb

PR #147, opened 2022-06-15
changed str in order to avoid confusions and collisions with Python

zaaachos:
Οverriding reserved words with your own will lead to confusion.

I will recommend to avoid using built-in names of Python, in order to avoid any collisions. 

so if you try the example below you will observe that local name overshadow the built-in str, rendering them un-usable.

image

Great work @lucidrains ! I would like to contribute more on this code


lucidrains:
@zaaachos duly noted :)

zaaachos:
Thanks in advance! Awesome work

PR #166, opened 2022-06-24
add sampels-seen and ema decay

nousr:


nousr:
wait a sec, lemme fix something (will update)

nousr:
okay, this is better

Issue #154, opened 2022-06-20
Automatic Model Checkpoint Push


nousr:
A utility that will automatically push a model's EMA weights to a given endpoint when a Trainers `.save` is called could be really helpful for increasing the autonomy and decreasing reliance on wandb.

This could take on the form of the Tracker's abstract wrapper where you might have a `LocalSave`, `WandbSave`, `HuggingFaceSave`, etc.

This might be a _good first issue_ if someone wants to jump in and tackle it?

Veldrovive:
This would be a great feature to have. Basically what it comes down to is making trackers more flexible. At the moment you can specify where to load from separately from where logs and checkpoints are stored, but logs and checkpoints must be saved in the same place. Once save, load, and log can all push to use different data locations it will be good.

In order to do this, I would suggest using the same pattern that is used for loading. Instead of having a separate save method for each tracker, have only one on the base class which uses some key from the config to specify the source and the rest of the keys in that section as kwargs.

It might also be nice to allow a user to pass a custom save and load function when creating the tracker so that if they want to house their checkpoints somewhere else they can do so. Maybe if you specify the save source as `custom` you could then substitute your own method in the initializer.

PR #172, opened 2022-06-26
Overhauled the tracker system

Veldrovive:
This is a response to #154. The existing tracker system turned out to be too restrictive for some features @nousr and I wanted. Mainly, we wanted to be able to log to W&B, but save to both W&B and huggingface (and have the models going to huggingface only include the prior/decoder and not the surrounding checkpoint state).

Changes:
The logging and saving systems have been separated.
Trackers are now built out of a logger, loader, and set of savers each of which is a class that defines its configuration in its constructor.
Moved tracker creation into the configuration class so that it is way easier to create a tracker outside of the `train_decoder` script.

There were a few breaking changes to the tracker api since some logic that used to be handled outside the tracker is now handled internally so more can be dependent on configuration options. Any places those caused issues were fixed.

I also wanted to handle the case where a somebody wants to use a logger, loader, or saver outside of the ones included in the repo. This can now be easily done by implementing the base class for the given handler, but you still have to manually insert that into the `tracker.py` file. A plugin system could be used instead, but that felt outside the scope of this refactor.

Todo:
- [ ] Add a plugin system so custom loggers, loaders, and savers can be used?

### DIFF
--- a/configs/README.md
+++ b/configs/README.md
@@ -91,21 +91,85 @@ Each metric can be enabled by setting its configuration. The configuration keys 
 
 **<ins>Tracker</ins>:**
 
-Selects which tracker to use and configures it.
+Selects how the experiment will be tracked.
 | Option | Required | Default | Description |
 | ------ | -------- | ------- | ----------- |
-| `tracker_type` | No | `console` | Which tracker to use. Currently accepts `console` or `wandb`. |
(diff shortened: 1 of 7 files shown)


lucidrains:
great PR! :100: :1st_place_medal:

lucidrains:
lgtm!

PR #175, opened 2022-06-26
bug fixes for text conditioning update

nousr:
found some bugs while starting a text conditioned run

@Veldrovive can you take a look as well and ensure this is correct?


Veldrovive:
Seems good to me. A change in my upsampler training branch does break the assumption that the the decoder should error when the first unet is text conditioned and text encodings are not passed in, but I can handle that separately.

PR #186, opened 2022-07-05
Add deepspeed support for decoder training

Veldrovive:
In order to train decoders of the size openAI uses, we need the optimizations deepspeed offers. This pull request changes initialization of the trainer to abide by what deepspeed requests.

Functionality:
- [x] Zero
- [ ] Text conditioning support
- [x] fp16

Getting fp16 working is our main concern at the moment as it is required to train larger models. Unfortunatly, when it is enabled, we get overflow warnings and infinities start appearing. The logging that appears is below.

Without anomaly detection enabled: [logs](https://pastebin.com/q5ua63Rs)

With anomaly detection enabled: [logs](https://pastebin.com/2aUYqGhe)

We are exploring solutions, but if you have any insight it would be appreciated. I was thinking it might have to do with https://github.com/pytorch/pytorch/issues/41527, but I did not see any difference when varying eps.

### DIFF
--- a/dalle2_pytorch/dataloaders/decoder_loader.py
+++ b/dalle2_pytorch/dataloaders/decoder_loader.py
@@ -1,6 +1,7 @@
 import os
 import webdataset as wds
 import torch
+from torch.utils.data import DataLoader
 import numpy as np
 import fsspec
 import shutil
@@ -255,7 +256,7 @@ def create_image_embedding_dataloader(
     )
     if shuffle_num is not None and shuffle_num > 0:
         ds.shuffle(1000)
-    return wds.WebLoader(
+    return DataLoader(
         ds,
         num_workers=num_workers,
         batch_size=batch_size,
--- a/dalle2_pytorch/trainer.py
+++ b/dalle2_pytorch/trainer.py
@@ -16,7 +16,7 @@ from packaging import version
 
 from ema_pytorch import EMA
 
-from accelerate import Accelerator
+from accelerate import Accelerator, DistributedType
 
 import numpy as np
 
@@ -424,6 +424,7 @@ class DecoderTrainer(nn.Module):
         self,
         decoder,
         accelerator = None,
+        dataloaders = None,
         use_ema = True,
         lr = 1e-4,
         wd = 1e-2,
@@ -476,8 +477,10 @@ class DecoderTrainer(nn.Module):
 
         self.register_buffer('step', torch.tensor([0.]))
 
-        decoder, *optimizers = list(self.accelerator.prepare(decoder, *optimizers))
+        decoder, train_loader, val_loader, *optimizers = list(self.accelerator.prepare(decoder, dataloaders["train"], dataloaders["val"], *optimizers))
 
+        self.train_loader = train_loader
+        self.val_loader = val_loader
         self.decoder = decoder
 
         for opt_ind, optimizer in zip(range(len(optimizers)), optimizers):
@@ -606,9 +609,12 @@ class DecoderTrainer(nn.Module):
 
         total_loss = 0.
 
+        using_deepspeed = self.accelerator.distributed_type == DistributedType.DEEPSPEED
+        using_amp = self.accelerator.mixed_precision != 'no'
+
         for chunk_size_frac, (chunked_args, chunked_kwargs) in split_args_and_kwargs(*args, split_size = max_batch_size, **kwargs):
-            # with autocast(enabled = self.amp):
-            with self.accelerator.autocast():
+            context_manager = autocast(enabled = using_amp) if using_deepspeed else self.accelerator.autocast()  # This version of accelerate does not handle deepspeed autocast correctly
+            with context_manager:
                 loss = self.decoder(*chunked_args, unet_number = unet_number, **chunked_kwargs)
                 loss = loss * chunk_size_frac
 
--- a/train_decoder.py
+++ b/train_decoder.py
@@ -279,6 +279,7 @@ def train(
     trainer = DecoderTrainer(
         decoder=decoder,
         accelerator=accelerator,
+        dataloaders=dataloaders,
         **kwargs
     )
 
@@ -326,7 +327,7 @@ def train(
         last_snapshot = sample
 
         if next_task == 'train':
-            for i, (img, emb, txt) in enumerate(dataloaders["train"]):
+            for i, (img, emb, txt) in enumerate(trainer.train_loader):
                 # We want to count the total number of samples across all processes
                 sample_length_tensor[0] = len(img)
                 all_samples = accelerator.gather(sample_length_tensor)  # TODO: accelerator.reduce is broken when this was written. If it is fixed replace this.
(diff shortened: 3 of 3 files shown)


lucidrains:
@Veldrovive Hi Aidan, and glad to see you have graduated to Deepspeed already :smile: 

so overflows are sort of routine in transformer training, and there are number of ways to alleviate the issue

for starters though, if using deepspeed, try giving the network a warmup of about 10k steps (it should be built into deepspeed)

the other thing that can be tried within the deepspeed framework is gradient clipping, and a value of 1.0 is good for starters

if all else fails, we'll have to figure out where the overflow is occurring, and if it is in the attention (as it often is), i can build in a few more countermeasures

lucidrains:
@Veldrovive for now, i've also removed the use of the functional `layer_norm`, just so we can rule out any potential issues there

lucidrains:
@Veldrovive are the fp16 results above in the context of deepspeed, or just regular DDP + fp16?

Veldrovive:
@lucidrains This is in the deepspeed context. I thought that fp16 might have been causing the NaNs we were seeing before, but I could replicate that in some short tests.

Removing `layer_norm` seems to have no effect. The integration with deepspeed is incomplete and things like LR schedulers or optimizer choice are not available yet. Just the basic optimizations. That being said, I can still do the poor man's learning rate scheduling. I'll start up a test run and update with results.

lucidrains:
@Veldrovive rest up Aidan! appreciate all the effort you have placed into the project so far and you are not obligated to push yourself, especially given a recent surgery :pray:

lucidrains:
> 

yup, i expect deepspeed not to be smooth sailing, as huggingface still has the "experimental" tag next to it, but if f16 with DDP works, then that rules out any issues with half precision!

rom1504:
deepspeed fp16 is true fp16 while ddp one is mixed precision
so deepspeed saves much more vram but also it's more risky
we had observed similar things at dalle-pytorch

Veldrovive:
The relevant [documentation](https://www.deepspeed.ai/docs/config-json/#fp16-training-options) for deepspeed and the explanation of the [referenced O2 amp configuration](https://nvidia.github.io/apex/amp.html#o2-almost-fp16-mixed-precision) that is similar to what fp16 does.

So basically some layer is being cast to fp16 that is blacklisted by the normal amp and the lack of precision is causing problems. Am I reading that correctly?

lucidrains:
@Veldrovive yea, in general fp16 training is more difficult, especially when there is attention and transformers in the network

i added a few tricks https://github.com/lucidrains/DALLE2-pytorch/commit/b9a908ff7522d52295dbdfd3487c9122e6cd4b98 that should help a bit for the attention / layernorms, however, i lack experience training u-nets, and not sure how they behave in fp16 setting

lucidrains:
i have a feeling warmups should help a bit, if it isn't some more obvious bug

we should also be open to the possibility that accelerate's deepspeed integration just isn't ready yet

Veldrovive:
Oh, speaking of which here's the [log](https://pastebin.com/L9JUyGZk) of a run done on a much lower learning rate. Around `1e-6`.

The integration does not especially matter here. In my test runs accelerate is doing very little besides call deepspeed initialize. I imagine we would have these same issues just using deepspeed.

lucidrains:
all in all, i'm not that worried because the Imagen paper shows that we don't really need to scale the unets to get good results. this is unlike what is going on for large language models

lucidrains:
@Veldrovive ohh ok, then it may just be the usual instabilities of training anything with attention

lucidrains:
@Veldrovive taking a shot in the dark haha https://github.com/lucidrains/DALLE2-pytorch/commit/1e4bb2bafb68ea24c26b3d5e170a900e01682489 do you have the full error message?

Veldrovive:
Full traceback with newest changes https://pastebin.com/5trCcHiH

lucidrains:
@Veldrovive https://github.com/lucidrains/DALLE2-pytorch/commit/f33453df9f04434369ef4474295ddb82ad1bab43 lol

lucidrains:
honestly, scaling with deepspeed probably won't be that critical for ddpm models as it is for language models

the Imagen paper shows that most of the work is done by the text encoder (and in DALL-E2's case, that would be CLIP)

the u-nets realistically won't be scaled beyond 1-2 billion parameters imo

Veldrovive:
Yea, I've really got to stop. But this project is exciting and I'm so used to pushing myself for uni that it only feels natural.

That last commit seems to have made it not error at the cost of the loss now immediately being `inf`. So that's neat. Anomaly detections says [...](https://pastebin.com/CZDxcWZs)

I'm not so convinced that scaling won't be necessary. It is also possible that CLIP isn't doing the same heavy lifting T5 is so the decoder needs to compensate. Deepspeed would also be helpful for just the increased speed it can provide.

Ok, I'm going to hit the hay early.

lucidrains:
@Veldrovive ok go rest, there's plenty of time, we'll figure it out :smile: 

that error is related to the learned variance, i'll take a look at it tomorrow morning

in the worst case we can just turn that off, since Imagen never used it

lucidrains:
https://github.com/lucidrains/DALLE2-pytorch/commit/1bd8a7835ab66f266daf11cff2aa56ed6a46a001 another shot in the dark

Veldrovive:
That shot seems to have hit something. I have a test run going with ZeRO  level 2 and fp16 enabled that has not error after around 2 million samples which is magnitudes better than before. I am not certain it is completely stable because the loss scaling seems to still be dropping at some points.

I also added support for bf16, but that training seems to be inefficient. I see some talk of similar things over here https://github.com/Lightning-AI/lightning/issues/11933, but if fp16 is stable that is not important.

lucidrains:
machine_learning_2x

Veldrovive:
Some info that was kind of expected, but useful anyway. Learned variance does not work with this fix. Also not a priority though.

lucidrains:
> Some info that was kind of expected, but useful anyway. Learned variance does not work with this fix. Also not a priority though.

ohh so the successful run with deepspeed isn't using learned variance?

Veldrovive:
Right, learned variance has to be off or there is always a loss overflow.

lucidrains:
> Right, learned variance has to be off or there is always a loss overflow.

ohhh got it, thanks! i'll try to debug it with a local version at https://github.com/lucidrains/denoising-diffusion-pytorch

lucidrains:
i'm almost tempted to just deviate from the dalle2 paper; remove the learned variance, adopt Karras' elucidating ddpm paper, call it a day. people are starting to see really good results with that over at Imagen-pytorch

Veldrovive:
Yea, honestly the interesting part of dalle2 is using both text and image embeddings such that you can do interesting operations in the image latent space. Outside of that I just keep seeing reasons why the dalle2 architecture in particular has redundant or unnecessary features. I haven't read Karras' ddpm paper, but I would be interested in seeing how other diffusion architectures work with the same prior.

Veldrovive:
Unfortunately, accelerate currently has a bug that prevents us from launching multi node training with deepspeed. The accelerate team is aware of the issue and we can track it here https://github.com/huggingface/accelerate/issues/498.

In the meantime if we want to train large models we will have to figure out how to use the deepspeed launcher to run the script instead of mpi and accelerate. This pull request has gotten very messy though and I made a new branch to condense all the tests down into one commit to work on further. This discussion is at an end so I am going to close the pull request and open a new one.

PR #191, opened 2022-07-08
Add the ability to auto restart the last run when started after a crash

Veldrovive:
We are finding on the HPC that it is common for runs to crash and be auto-started again. This pull requests allows that auto restart to resume the previous run and load a checkpoint only when the run was auto restarted.

It also fixes a variable naming issue and unnecessary accelerator prepare call that was causing issues.

### DIFF
--- a/dalle2_pytorch/trackers.py
+++ b/dalle2_pytorch/trackers.py
@@ -66,7 +66,7 @@ class BaseLogger:
 
     def get_resume_data(self, **kwargs) -> dict:
         """
-        Sets the config arguments that along with { "resume": True } will be used to resume training.
+        Sets tracker attributes that along with { "resume": True } will be used to resume training.
         It is assumed that after init is called this data will be complete.
         If the logger does not have any resume functionality, it should return an empty dict.
         """
@@ -164,9 +164,9 @@ class WandbLogger(BaseLogger):
     def get_resume_data(self, **kwargs) -> dict:
         # In order to resume, we need wandb_entity, wandb_project, and wandb_run_id
         return {
-            "wanb_entity": self.entity,
-            "wandb_project": self.project,
-            "wandb_run_id": self.wandb.run.id
+            "entity": self.entity,
+            "project": self.project,
+            "run_id": self.wandb.run.id
         }
 
 logger_type_map = {
--- a/dalle2_pytorch/trackers.py
+++ b/dalle2_pytorch/trackers.py
@@ -420,30 +420,29 @@ class Tracker:
         self.dummy_mode = dummy_mode
 
     def _load_auto_resume(self) -> bool:
-        # First, we check if data_path / auto_resume.json exists
-        if self.auto_resume_path.exists():
-            # First, we check if the logger has autoresume enabled. If not, we remove the autoresume file and move on
-            if not self.logger.auto_resume:
-                print(f'Removing auto_resume.json because auto_resume is not enabled in the config')
-                self.auto_resume_path.unlink()
-                return False
-            else:
-                # Otherwise we read the json into a dictionary will will override parts of logger.__dict__
-                with open(self.auto_resume_path, 'r') as f:
-                    auto_resume_dict = json.load(f)
-                # Check if the logger is of the same type as the autoresume save
-                if auto_resume_dict['logger_type'] != self.logger.__class__.__name__:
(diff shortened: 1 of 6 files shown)


train_decoder.py
```diff
@@ -289,9 +289,9 @@ def train(
     sample = 0
     samples_seen = 0
     val_sample = 0
-    step = lambda: int(trainer.step.item())
+    step = lambda: int(trainer.steps[0].item())
```
lucidrains, line 292: offered a helper method https://github.com/lucidrains/DALLE2-pytorch/commit/d7bc5fbedd85564f7050534d06a3d94da50e6c91#diff-617450527162fa367141dbf45e8b201673573479820af0ffc56ba93b7f70947fR537 this would only retrieve the number of training steps for the first unet

dalle2_pytorch/trackers.py
```diff
@@ -398,20 +419,65 @@ def __init__(self, data_path: Optional[str] = DEFAULT_DATA_PATH, overwrite_data_
...
+                # Check if the logger is of the same type as the autoresume save
+                if auto_resume_dict['logger_type'] != self.logger.__class__.__name__:
+                    raise Exception(f'The logger type in the auto_resume file is {auto_resume_dict["logger_type"]} but the current logger is {self.logger.__class__.__name__}. Either use the original logger type, set `auto_resume` to `False`, or delete your existing tracker-data folder.')
+                # Then we are ready to override the logger with the autoresume save
+                self.logger.__dict__["resume"] = True
+                print(f"Upading {self.logger.__dict__} with {auto_resume_dict}")
+                self.logger.__dict__.update(auto_resume_dict)
+
+                return True
+        else:
```
lucidrains, line 443: more a stylistic critique, but one way for a cleaner function when you have all these branching nested if / else, is to put the edge cases near the top and do an early return

so in your function here

```python
if not self.auto_resume_path.exists():
  # log
  return False

if not self.logger.auto_resume:
  # log
  return False

# rest of the main logic loading from json + logging
return True
```

that way it is more readable and the edge cases are organized near the top

dalle2_pytorch/trackers.py
```diff
@@ -398,20 +419,65 @@ def __init__(self, data_path: Optional[str] = DEFAULT_DATA_PATH, overwrite_data_
...
+            else:
+                # Otherwise we read the json into a dictionary will will override parts of logger.__dict__
+                with open(self.auto_resume_path, 'r') as f:
+                    auto_resume_dict = json.load(f)
+                # Check if the logger is of the same type as the autoresume save
+                if auto_resume_dict['logger_type'] != self.logger.__class__.__name__:
+                    raise Exception(f'The logger type in the auto_resume file is {auto_resume_dict["logger_type"]} but the current logger is {self.logger.__class__.__name__}. Either use the original logger type, set `auto_resume` to `False`, or delete your existing tracker-data folder.')
+                # Then we are ready to override the logger with the autoresume save
+                self.logger.__dict__["resume"] = True
+                print(f"Upading {self.logger.__dict__} with {auto_resume_dict}")
```
lucidrains, line 439: small typo

lucidrains:
lgtm save for the training step being always fetched from the first unet! :100:

train_decoder.py
Veldrovive: Will change to using that. I will also have to change the upsampling trainer pr to take advantage of this.

unet 0 is selected arbitrarily for no other reason than it does not matter which unet is chosen and 0 is always present. I am actually just going to remove this variable entirely and replace calls that use it with the helper function since that will be more reliably correct. It will still choose unet 0 though. I could also allow this to be configured, but since it only effects logging that does not seem neccesary.

lucidrains: oh ok, as long as you're aware!

Veldrovive:
Made the changes and retested. Ignore the spelling mistake in the commit message.

train_decoder.py
Veldrovive: I do wish there was a more satisfying way to do this, but it really is just an arbitrary choice and as long as the index is consistent between runs it is fine. The way it is handled in the upsampler training is that it takes the first training unet for the log step.

I guess we could go back to the original system of tracking the step separately from the unet step, but unet step is more informative so that doesn't seem like a good idea.

PR #193, opened 2022-07-08
setup simple ci

rom1504:


rom1504:
It runs the decoder training with very small settings

Will make sure things keep working

PR #201, opened 2022-07-10
Add another upsample technique

MalumaDev:
Add another upsample technique: PixelShuffle (https://arxiv.org/pdf/1609.05158v2.pdf).
This technique should help to avoid the emergence of checkerboard artifacts.

### DIFF
--- a/dalle2_pytorch/dalle2_pytorch.py
+++ b/dalle2_pytorch/dalle2_pytorch.py
@@ -1,5 +1,7 @@
 import math
 import random
+from enum import Enum
+
 from tqdm.auto import tqdm
 from functools import partial, wraps
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


lucidrains:
@MalumaDev thank you 🙏 ! does this fix the upsampler issues for you?

MalumaDev:
I have yet to test it for full training, but I have used it in other works. I have only one GPU and the test will require some times.

lucidrains:
Cool, I'll test it next week trying to debug the artifacts we see in the other issue

lucidrains:
@marunine if Katherine is doing it, it is probably a good idea haha https://github.com/lucidrains/DALLE2-pytorch/commit/bdd62c24b367847abeb1bbf5f2b7b77311ed84fb

marunine:
I remembered reading about it from one of the Adam authors before - comes from the GLOW paper.

The good news is that I'm no longer seeing checkerboarding on the Imagen repo with the zero initialization + pixel shuffle at least. Haven't tried out the zero initialization alone, but no checkerboarding on a fresh run even in the early steps is great.

lucidrains:
@marunine @MalumaDev  thank you both!

PR #202, opened 2022-07-10
Show Unet parameter count at beginning

Veldrovive:
This is a feature from the upsampler training branch that is useful enough to add here. If clip is included the parameter count of the decoder is higher than the parameter count of the unets. With this it now shows both.


rom1504:
Nice

Issue #203, opened 2022-07-12
Error related to text mask 


rom1504:
```
File "/fsx/dalle2/.dalle_env_38/lib64/python3.8/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
losses = self.p_losses(unet, image, times, image_embed = image_embed, text_encodings = text_encodings, text_mask = text_mask, lowres_cond_img = lowres_cond_img, predict_x_start = predict_x_start, learned_variance = learned_variance, is_latent_diffusion = is_latent_diffusion, noise_scheduler = noise_scheduler)
File "/fsx/dalle2/DALLE2-pytorch/dalle2_pytorch/dalle2_pytorch.py", line 2340, in p_losses
return forward_call(*input, **kwargs)
File "/fsx/dalle2/DALLE2-pytorch/dalle2_pytorch/dalle2_pytorch.py", line 1817, in forward
model_output = unet(
File "/fsx/dalle2/.dalle_env_38/lib64/python3.8/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
text_keep_mask = text_mask & text_keep_mask
RuntimeError: The size of tensor a (0) must match the size of tensor b (26) at non-singleton dimension 0
return forward_call(*input, **kwargs)
File "/fsx/dalle2/DALLE2-pytorch/dalle2_pytorch/dalle2_pytorch.py", line 1817, in forward
text_keep_mask = text_mask & text_keep_mask
RuntimeError: The size of tensor a (0) must match the size of tensor b (26) at non-singleton dimension 0
```

Recently introduced but not sure which commit

lucidrains:
@rom1504 oh oops, might be me :cry:  https://github.com/lucidrains/DALLE2-pytorch/commit/e0835acca97ba75b226d6eb2023200a293929fbb may or may not fix it haha

rom1504:
```
Traceback (most recent call last):
  File "/fsx/dalle2/upsampler/train_decoder.py", line 610, in <module>
    main()
  File "/fsx/dalle2/.upsampler/lib64/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/fsx/dalle2/.upsampler/lib64/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/fsx/dalle2/.upsampler/lib64/python3.8/site-packages/click/core.py", line 1404, in invoke
[...]

PR #204, opened 2022-07-12
add test for upsampler training

rom1504:
should reproduce #203 

(which likely fails because #181 is not done)


lucidrains:
it is working

PR #205, opened 2022-07-13
Fixed issue with dataloaders not loading text

Veldrovive:
I changed the decoder dataloaders to be prepared by accelerate when I added deepspeed. Unfortunately, accelerate removes the ability to collate strings, likely since it cannot gather strings. We are not using the distributed dataloader features, however, so I removed this.

### DIFF
--- a/train_decoder.py
+++ b/train_decoder.py
@@ -323,7 +323,7 @@ def train(
         last_snapshot = sample
 
         if next_task == 'train':
-            for i, (img, emb, txt) in enumerate(trainer.train_loader):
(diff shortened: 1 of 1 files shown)


lucidrains:
@Veldrovive  wow, that was quick! thanks Aidan!

muellerzr:
@lucidrains and @Veldrovive I've been thinking about this since I saw this PR and comment. Would a utility that converts the string to a long tensor of shape `1xnxchar_limit` work to solve this issue?

E.g. is this occurring at the dataset level where we could write a utility you could integrate into your custom pipeline you have going on?

PR #211, opened 2022-07-21
Prior updates

nousr:
This PR makes the following notable changes:
* adopt full accelerator support in the diffusion prior trainer
* update prior training script to leverage the latest tracker version
    * can resume interrupted epoch
    * leverages new saving features of tracker
* trainer now utilizes a short warmup period to stabilize training
* updates to training script to mirror trainer's changes
    * ability to evaluate on multiple time-steps
    * save and evaluate on time-based intervals (rather than steps

### DIFF
--- a/configs/train_prior_config.example.json
+++ b/configs/train_prior_config.example.json
@@ -1,18 +1,14 @@
 {
     "prior": {
         "clip": {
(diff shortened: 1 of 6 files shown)


lucidrains:
:100: :rocket:

PR #234, opened 2022-08-20
Fixed issues with clip and deepspeed fp16

Veldrovive:
When using deepspeed fp16 mode clip threw errors because the libraries used were not intended for that purpose. To solve this, we now exclude clip from the call to the deepspeed prepare. This has some wider ramifications on the decoder training script because it means the clip model must be passed around in parallel with the decoder model since we can no longer reliably keep clip inside the decoder.

Todo:
- [x] Finish the more complete test (This change has broader effects than would be caught by the ghActions test)

### DIFF
--- a/dalle2_pytorch/train_configs.py
+++ b/dalle2_pytorch/train_configs.py
@@ -241,7 +241,7 @@ class DecoderConfig(BaseModel):
     clip: Optional[AdapterConfig]   # The clip model to use if embeddings are not provided
     channels: int = 3
     timesteps: int = 1000
-    sample_timesteps: Optional[SingularOrIterable[int]] = None
+    sample_timesteps: Optional[SingularOrIterable[Optional[int]]] = None
     loss_type: str = 'l2'
     beta_schedule: ListOrTuple[str] = None  # None means all cosine
     learned_variance: SingularOrIterable[bool] = True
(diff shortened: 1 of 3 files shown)


Veldrovive:
As expected there was a small problem with typecasting because of the recent addition of the softmax. This will error on the einsum since attn and v may have different dtypes. https://github.com/lucidrains/DALLE2-pytorch/blob/7762edd0ff45788ed45c54c96fb080b28522a082/dalle2_pytorch/dalle2_pytorch.py#L881-L886

It's easy enough to fix by just casting back to the dtype of `sim` after, but I don't know what the intention of the fp32 softmax is so I want to make sure it shouldn't also be fp32 for some other manipulation. Could you take a look @lucidrains?

lucidrains:
@Veldrovive oh! didn't know it behaved that way

https://github.com/lucidrains/DALLE2-pytorch/commit/083508ff8e8861f1b1670e807bda25b538ea032c should fix it :crossed_fingers:

Veldrovive:
Yep, all fixed now. Everything appears to be working now.

Veldrovive:
[The test run](https://wandb.ai/veldrovive/dalle2_train_decoder/runs/enpj9x7c?workspace=user-veldrovive) shows similar progress to past runs.

PR #242, opened 2022-09-15
Minor Decoder Train Script Fixes

nousr:
I was testing out some stuff and ran into some errors.
- tokenization device placement
- lpips image distribution (due to floating point errors)

@Veldrovive if you can just double check that this looks good before we merge 🤝


Veldrovive (approved):
All changes look good to me.

PR #254, opened 2022-10-19
Fix potential dtype conversion bug

JiaHeng-DLUT:
The return value will be converted to 0 when x is an integer data type.

### DIFF
--- a/dalle2_pytorch/dalle2_pytorch.py
+++ b/dalle2_pytorch/dalle2_pytorch.py
@@ -1542,7 +1542,7 @@ class SinusoidalPosEmb(nn.Module):
         emb = math.log(10000) / (half_dim - 1)
         emb = torch.exp(torch.arange(half_dim, device = device, dtype = dtype) * -emb)
         emb = rearrange(x, 'i -> i 1') * rearrange(emb, 'j -> 1 j')
-        return torch.cat((emb.sin(), emb.cos()), dim = -1).type(dtype)
+        return torch.cat((emb.sin(), emb.cos()), dim = -1)
 
 class Block(nn.Module):
     def __init__(


lucidrains:
@JiaHeng-DLUT ohh, is there a place where it would be an int type? i think it is type converted to be float here https://github.com/lucidrains/DALLE2-pytorch/blob/main/dalle2_pytorch/dalle2_pytorch.py#L2155

JiaHeng-DLUT:
Yes. Here is the traceback:
https://github.com/lucidrains/DALLE2-pytorch/blob/HEAD/dalle2_pytorch/dalle2_pytorch.py#L1098
https://github.com/lucidrains/DALLE2-pytorch/blob/HEAD/dalle2_pytorch/dalle2_pytorch.py#L1355
https://github.com/lucidrains/DALLE2-pytorch/blob/HEAD/dalle2_pytorch/dalle2_pytorch.py#L1463
https://github.com/lucidrains/DALLE2-pytorch/blob/HEAD/dalle2_pytorch/dalle2_pytorch.py#L1455

There is a conflict at https://github.com/lucidrains/DALLE2-pytorch/blob/HEAD/dalle2_pytorch/dalle2_pytorch.py#L972.
`nn.Embedding` needs the input to be int or long, while `SinusoidalPosEmb` needs the input to be float. I recommend removing the explicit data type conversion to support int. Also, the type will be converted implicitly when the input is float.

lucidrains:
@JiaHeng-DLUT yes you are right! :pray: this was a big bug actually, and should be resolved in https://github.com/lucidrains/DALLE2-pytorch/commit/41fabf29220d0469af9c2681068e2ef99caa0085 thank you for uncovering this!

Issue #305, opened 2023-09-30
Field missing in config.


u1ug:
Got error Field required [type=missing, input_value={'data_path': '.prior1, ..../best_checkpoint.pth'}]}, input_type=dict] while running train_diffusion_prior.py

PR #306, opened 2023-10-04
Make TrackerLoadConfig optional

neilkimn:
Fixes https://github.com/lucidrains/DALLE2-pytorch/issues/305


lucidrains:
@neilkimn thanks for the quick fix

